Issue #3, opened 2020-10-21
Use current path for Icon creation


Mimoja:
https://github.com/decentespresso/de1app/blob/main/de1beta/gui.tcl#L577

Mimoja:
fixed with #4

PR #4, opened 2020-10-21
gui: Make app icon creation folder independant

Mimoja:
So far the path to the de1app sourcefolder had to be placed directly under /sdcard.
This leads to the de1app repository beeing not directly executable. While the app would run
the icon creation would not.
To fix that we are querring the current scripts folder to find the correct path from there.


Mimoja:
This fixes #3

decentjohn:
What you've done is exactly what I intended to eventually implement as well.  Perfect.

PR #13, opened 2020-10-24
comms: Cleanup connection managment to allow for use with different connection modes

Mimoja:
comms: Cleanup connection managment to allow for use with different connection modes

So far we were only able to connect to the DE1 via BLE. To be able to remove this
restriction we need not only to safe the connectivity mode as well as the but build
generic versions of connection managment functions.
In this commit we also add the ability to search for USB devices and connect to them.

### DIFF
--- a/de1plus/de1_comms.tcl
+++ b/de1plus/de1_comms.tcl
@@ -1,5 +1,7 @@
 package provide de1_comms 1.0
+package require de1_machine
 package require de1_bluetooth
+package require de1_usb
 
 ### Globals
 set ::failed_attempt_count_connecting_to_de1 0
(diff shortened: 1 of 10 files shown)


de1plus/de1_comms.tcl
```diff
@@ -138,97 +140,109 @@ proc append_to_de1_list {address name type} {
 	msg "Scan found DE1: $address"
 	set ::de1_device_list $newlist
 	catch {
-		fill_ble_listbox
+		fill_de1_listbox
 	}
 }
 
-proc append_to_de1_list {address name type} {
```
Mimoja, line 145: How did we manage to end up with the same function multiple times?
Maybe patch files are rightfully not the future :D

decentjohn: It can be a user error, simply copy/paste. I've made this error in the past many times.

Tcl allows you to redefine a proc{} at any time, without a warning.

decentjohn:
If the two procs are different, make sure you remove the earlier one, as the later defined one is the one that has been in production.

decentjohn (approved):
it's a bit too much code to review on a web page.  It'll have to be tested to see if it works and if any bugs arise.

decentjohn:
Am I supposed to do something to finalize this Issue?

Mimoja:
lets just keep it open for the moment. I will add to this once i release the rest of the USB thingy.

decentjohn:
Did you want this reviewed now by me, or delay review?

Mimoja:
Not worth your time right now

PR #16, opened 2020-11-01
Hide risque skins behind feature toggle

Mimoja:

grafik

if enabled

grafik


decentjohn:
I mentioned a few changes I'd like to see on this before we make it available to everyone:

1) change to "hide risque skins"

2) move this checkbox setting to the skins page, checking it cause re-load of the listbox.

3) don't show the setting at all, if there are no risque skins in the installation

4) check it on by default

https://3.basecamp.com/3671212/buckets/7351439/messages/3168343099

Mimoja:
Given the feedback on the forums I am no longer inclined adding the skins to the app. Closing for now.

Issue #17, opened 2020-11-09
ble listboxes need scrollbars


decentjohn:
At the korea coffee show, there were so many acaia, felicita and other scales, that Skale was never in the top few to pick from.  

So... we need to add scrollbars to the ble pick lists, for this odd case.

Mimoja:
lol. amazing :3

barneyhawes:
John - thanks for bring in these changes. You can close this issue now.
FYI - it would have closed automatically if you'd accepted the pull request instead of duplicating the changes. Between us we'll figure out the best ways to use Git!

PR #20, opened 2020-11-19
Ble scrollbar

barneyhawes:
Fix for #17 
This fix adds scrollbars for both bluetooth device lists, for consistency.
I also created a single function for sizing scrollbar boxes to reduce code duplication.
I have not tested with real bluetooth devices so we need to confirm that the slight reduction in listbox width does not prevent the whole address being displayed.

### DIFF
--- a/de1plus/skins/default/de1_skin_settings.tcl
+++ b/de1plus/skins/default/de1_skin_settings.tcl
@@ -74,6 +74,13 @@ set slider_trough_color #EAEAEA
 set chart_background_color #F8F8F8
 ##############################################################################################################################################################################################################################################################################
 
+proc set_scrollbar_dimensions { scrollbar_widget listbox_widget } {
+		# set the height of the scrollbar to be the same as the listbox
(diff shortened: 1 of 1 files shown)


decentjohn:
Ok I've reviewed this and made some small changes to listbox width & height adjustments for running this app under stock tcl.  

I'm not sure @barneyhawes if I need to get your OK to commit my changes back to your tree before accepting your changes.  

Afraid this might need to wait for @Mimoja as there are several branches now.

Screen Shot 2020-11-20 at 2 28 17 PM

decentjohn:
I'm trying to use github desktop, which has support for reviewing PRs and once I understand how to use it correctly, it'll be nice.

barneyhawes:
Let's see what the advice is for the best way for you to accept a pull request with your modifications applied.

In the meantime, if you want to send me a diff or the updated tcl files then I can update my branch.

decentjohn:
manually merged

PR #22, opened 2020-11-20
machine: add hidden setting to lock the screen if wanted

Mimoja:
We are seeing customers turning off their machines over night, not trusting the
power usage. To accomodate for their usecase we are disabeling the screenlock if
requested.

### DIFF
--- a/de1plus/machine.tcl
+++ b/de1plus/machine.tcl
@@ -418,6 +418,7 @@ array set ::settings {
 	force_acaia_heartbeat 0
 	comms_debugging 0
 	scale_stop_at_half_shot 0
+	lock_screen_during_screensaver 0
 }
 
 if {[de1plus]} {
@@ -855,6 +856,12 @@ proc start_water {} {
 proc start_idle {} {
 	msg "Tell DE1 to start to go IDLE (and stop whatever it is doing)"
 
+	# Ensure we are not locking the screen during use.
+	# This is only relevant when waking up the machine
+	if  {[sdltk screensaver] == 1} {
+		sdltk screensaver off
+	}
+
 	if {$::de1(device_handle) == 0} {
 		update_de1_state "$::de1_state(Idle)\x0"
 		ble_connect_to_de1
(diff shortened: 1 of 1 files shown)


barneyhawes:
Is there a reason to hide this setting? Sounds like it could be useful in other situations.

I press the tablet power button to turn off the screen once the app is sleeping. I then press the tablet power button to wake the tablet - and it would be great if this automatically disabled the screen saver. Maybe this is a separate FR? But I imagine people will have different ways of using it, so exposing settings like this might be good.

Mimoja:
@decentjohn HATES if i try to sneak in more settings to the misc screen. sooo.. Ehmm.. Lets just pretend I learned my lesson here
(and am working on documenting all internal settings for the wiki)

barneyhawes:
Well, I agree with the principle of having as few settings as possible, and I guess we can always expose something later.

decentjohn:
My only problem with powering the DE1 off overnight is if the tablet battery goes low.

There is logic in the app that check for low battery at app startup and warns them to not power off in the future.  

This feature seems to have resolved the main issue.

Separate idea, which I've researched, is on rooted tablets it's possible for the app to:
1) power off the tablet when USB power has been off for XX minutes
2) auto-launch the app when booting

this is something the android-on-cars hackers have figured out, it's quite nice.

But Android security improvements have made this harder.  There is an app for Android that can automatic this, and I've discussed it on Diaspora...

decentjohn:
I don't mind the idea of tablet power off == de1 sleep, though I can think of cases where the person wouldn't want that.  

I don't see any way for an app to know that the screen is off.  Maybe the brightness setting goes to zero?  Worth investigating?

Anyway, this doesn't seem like a priority to me.

decentjohn:
As far as a phantom screen press turning the app/de1 back on, I think the better solution is to use the facilties android provides, such as:
1) an adroid screen saver on top of the DE1 saver
2) restoring the screen auto-off feature on Android, so the screen goes off after X minutes

Mimoja:
Locking the screen in comparison to setting the Display to 0 and showing black has completely different effects. The PowerManager can clock down the cpu more aggressive, the GPU can be turned off, wifi and BLE can go to standby.
If the screensaver allows android to Lock the screen after a couple of seconds it will power down by a Lot and save a lot of energy. In my test the tablet can show the black screensaver for 18 hours, with screen locked I am at 30 hours and counting and still at 68% battery.
If having you machine plugged in over night is a deal breaker to some we can at least have the tablet survive it.

barneyhawes:
I power the tablet from a separate USB wall socket instead of from the DE1. Others may do the same - this could be recommended if you turn the DE1 off overnight.

Mimoja:
To quote from the forum:
```
Hello,
any update here on power down consumption?
For me the 4-5W are too high, this is not state of the art in electronics now, target is 0 or close to 0W.
```

PR #30, opened 2020-12-03
Add "keep awake" option for scheduler

lawtalker:
If the user adds `scheduler_logic 1` to settings.tdb, the following changes are implemented:

1. In vars.tcl, the flag `scheduler_awake` is set to 1 at the scheduled wake time, and unset at scheduled sleep time. At the scheduled sleep time, the machine does not immediately sleep (that instead is left for the idle timer to deal with).
2. In gui.tcl, if `scheduler_awake` is set to 1, don't sleep (or defer sleep for 60 seconds, to be more precise). Once `scheduler_awake` is unset, the machine will sleep per its regular settings.

Because these changes are implemented only if the user has chosen to edit settings.tdb, they hopefully are non-controversial.

Context: https://3.basecamp.com/3671212/buckets/7351439/messages/1529734776#__recording_3263004744

### DIFF
--- a/de1plus/gui.tcl
+++ b/de1plus/gui.tcl
@@ -927,7 +927,7 @@ proc show_going_to_sleep_page  {} {
 		return
 	}
 
-    if {[ifexists ::app_updating] == 1} {
+    	if {[ifexists ::app_updating] == 1} {
 		msg "delaying screen saver because tablet app is updating"
 		delay_screen_saver
 		return
@@ -937,7 +937,16 @@ proc show_going_to_sleep_page  {} {
 		msg "delaying screen saver because firmware is updating"
 		delay_screen_saver
 		return
-	}	
+	}
+	
+	# make scheduler work on "keep awake" logic
+	if {[ifexists ::scheduler_awake] == 1) {
+		msg "delaying screen saver 60 sec because of scheduler"
+		stop_screen_saver_timer
+                if {$::settings(screen_saver_delay) != 0 } {
+                        set ::screen_saver_alarm_handle [after 60000 "show_going_to_sleep_page"]
+		return
+	}
 
 	puts "show_going_to_sleep_page"
  	if {$::de1(current_context) == "sleep" || $::de1(current_context) == "saver"} {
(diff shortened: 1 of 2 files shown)


lawtalker:
This is my first ever pull request for the DE code. Let me know if there is a different protocol I should follow for making requests like this.

decentjohn:
Thanks for doing this work.

I want to do a poll of Diaspora users to determine if this should be the new default behavior.

decentjohn:
Diaspora poll
https://3.basecamp.com/3671212/buckets/7351439/messages/3269357872

lawtalker:
BTW, regarding the further commit I made earlier this morning: I just figured out how my machine got in a state where scheduler_awake was not set when the scheduler_sleep got called. This morning, I restarted the app before making my morning drink. That happened _after_ my scheduler_wake time, but _before_ my scheduler_sleep time. 

So when the app got to scheduler_sleep, scheduler_wake had never run, which meant there was nothing to unset. 

In any event, the -nocomplain option should suppress the error, which is exactly what we want.

This does make me realize that, in an ideal world, the app would check on startup whether it should be in "keep awake" mode, by seeing if the current time is between the wake and sleep times. 

But that's a corner case that doesn't need to be fixed right now. I can submit a pull request for that later.

lawtalker:
@decentjohn The poll results seem pretty definitive. Let me know if you'd like me to add further commits to make the new logic the default logic, and more generally if there is anything I can do to help get this patch merged in.

decentjohn:
Totally agree, it's just waiting on me to review your code for bugs. 🤪

lawtalker:
Hah. Fair enough. Just let me know if I can help.

decentjohn:
I liked the feature, but I wasn't crazy about how you implemented it.  I reimplemented the feature from scratch and committed it here: https://github.com/decentespresso/de1app/commit/c542909e3e6444abff22bf465a192c2cff3b7a3b

I'm fairly sure mine does the same thing, except that it's now the permanent new behavior, and not optional.

Issue #1, opened 2020-10-20
Listbox scroll bars have the wrong range


barneyhawes:
The lists of profiles, skins, etc don't scroll correctly. The bottom part of the scroll bar is not used.
See discussion on [Decent Diaspora (Basecamp)](https://3.basecamp.com/3671212/buckets/7351439/messages/2343554942).

Repro steps:
- Open settings
- Select Presets
- Select eye/magnifier icon to show whole list
- Drag the scroll bar down. The list reaches the bottom when the scroll bar is at around 70%

This also occurs in the list of Skins, and would happen in other lists which use scrolling.

The problem is the the amount by which the list can be scrolled should be the height of the items that are outside the visible part of the list, not the height of the entire list.

barneyhawes:
Created a Pull request (#2) with the fix for this.

barneyhawes:
Question for @Mimoja (or anyone else): now that this pull request has been accepted, what's the best thing to do with the branch on my fork? I'd like to "close" it but the closest option seems to be to delete it. I assume that will have no ill effects on the Decent code repo... but it would be good to get your opinion.

decentjohn:
Yeah, I think close is fine.

decentjohn:
With 14 items in the listbox, the Weber Spring Clean profile doesn't scroll to the bottom. Only 11 are displayed, and I have to use the mouse scroll wheel to show more.  Confirmed this same problem on the android tablet.


![Screen Shot 2020-12-08 at 10 15 59 AM](https://user-images.githubusercontent.com/11965250/101429634-60387380-393e-11eb-9713-f40b7523baee.jpg)
![Screen Shot 2020-12-08 at 10 15 55 AM](https://user-images.githubusercontent.com/11965250/101429641-64fd2780-393e-11eb-9d39-57923e2e0cfb.jpg)

barneyhawes:
Oh dear.  My first thought is that this might be a rounding error, but that probably wouldn't give an off-by-3 result.
I'm working right now but will take a look when I get a chance.

barneyhawes:
Further suggested improvements to scroll bar implementation:

1. Scroll bar position does not update correctly after adding/removing items
2. Set the scroll bar range and small/large step values according to the number of items. This would need to be updated each time the number of items in the list changes.
[...]

PR #32, opened 2020-12-10
Fixed list box scroll bar ranges

barneyhawes:
Fix for #1 where some scrollbars did not go from zero to one.

### DIFF
--- a/de1plus/skins/default/de1_skin_settings.tcl
+++ b/de1plus/skins/default/de1_skin_settings.tcl
@@ -290,7 +290,7 @@ add_de1_widget "settings_2c" listbox 70 310 {
 set ::advsteps_slider 0
 
 # draw the scrollbar off screen so that it gets resized and moved to the right place on the first draw
-set ::advsteps_scrollbar [add_de1_widget "settings_2c" scale 10000 1 {} -from 0 -to .50 -bigincrement 0.2 -background "#d3dbf3" -borderwidth 1 -showvalue 0 -resolution .01 -length [rescale_x_skin 400] -width [rescale_y_skin 150] -variable ::advsteps -font Helv_10_bold -sliderlength [rescale_x_skin 125] -relief flat -command {listbox_moveto $::advanced_shot_steps_widget $::advsteps_slider}  -foreground #FFFFFF -troughcolor "#f7f6fa" -borderwidth 0  -highlightthickness 0]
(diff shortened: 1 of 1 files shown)


decentjohn:
Was this your intended solution to this problem, namely to leave 50% of the bottom of the scroll bar empty when at the bottom of the list? It does solve the issue of not being able to scroll to the bottom, but it's kind of confusing, as it implies to the user that they're only 50% through the listbox choices.

Screen Shot 2020-12-10 at 5 42 14 PM

barneyhawes:
There's something odd here that I can't quite work out. When the scroll bar value changes, the `listbox_moveto_new` proc is called with the argument `$dest2` as the new value of the scroll bar.
This works correctly when holding the mouse buttons down and dragging: the value is in the range 0..1.
However, as soon as you release the mouse button, the function gets called one extra time, with the value capped at 0.5.

I haven't yet figured out why the function is being called a second time. In the above commit I changed the max value of this scroll bar from 0.5 to 0.1, so I think there might be somewhere else in the code that assumes the max value of the slider is 0.5.

By the way, I changed all the sliders to operate over the range 0..1. A better implementation would be to set the range to the actual number of items in the list, which would mean that the small increment would work correctly.

barneyhawes:
@decentjohn I'm stuck on this one at the moment. 

I added some debugging for the calls to `listbox_moveto_new`, `scale_scroll_new` and when the value of `::advsteps_slider` changes.

As you can see in [this log file](https://github.com/decentespresso/de1app/files/5672677/log.txt), this is working as expected when scrolling to the bottom of a list... until you get to the end and release the mouse.

- Adjusting the slider invokes the calls to `listbox_moveto_new`.
- When this reaches the point where it changes the top item for the list box, the list box is updated. This invokes a call to `scale_scroll_new`.
- The value of `::advsteps_slider` is then updated.

The final call to `listbox_moveto_new` is the problem.

For some reason there is this extra function call that sets the value to 0.5. 
This call occurs when the mouse button is released.
It only happens for the shot steps slider (none of the other sliders generate this extra function call).

Can you figure out why this particular scroll bar's command is being invoked an extra time with the wrong value when the mouse is released?

decentjohn:
I've got a theory as to what's going on.

In this command in the skin settings:

`set ::advsteps_scrollbar [add_de1_widget "settings_2c" scale 10000 1 {} -from 0 -to 1.0 -bigincrement 0.2 -background "#d3dbf3" -borderwidth 1 -showvalue 0 -resolution .01 -length [rescale_x_skin 400] -width [rescale_y_skin 150] -variable ::advsteps -font Helv_10_bold -sliderlength [rescale_x_skin 125] -relief flat -command {listbox_moveto $::advanced_shot_steps_widget $::advsteps_slider}  -foreground #FFFFFF -troughcolor "#f7f6fa" -borderwidth 0  -highlightthickness 0]`

I removed the variable reference to 
`-variable ::advsteps`

and this problem goes away.

I think this means the scroll bar will not longer auto-move itself when the selected step changes, but I'm not sure that would ever occur.

Not quite sure what other effects this change will have, as it's been a while since I wrote the scroll bar support, and I didn't get it 100% right either (ie, you fixed some bugs in it).

barneyhawes:
@decentjohn That's a good spot. I think this is actually a typo: the variable name should be `::advsteps_slider`. I'll do another pull request.

decentjohn:
ah ha! nicely spotted to you too.

PR #33, opened 2020-12-12
Add minimal webserver listening on 8080 that allows to toggle the power

Mimoja:
Tested on windows and on device for multiple hours and shots over a single day. Supports on / off.
Instead of toggle it us using dedicated endpoints to recude the risk of accidental use.

Interface is absolutely up for discussion together with everything else. Please dont merge without proper discussion. 

### DIFF
--- a/de1plus/machine.tcl
+++ b/de1plus/machine.tcl
@@ -418,6 +418,10 @@ array set ::settings {
 	comms_debugging 0
 	scale_stop_at_half_shot 0
 	lock_screen_during_screensaver 0
+	webserver_enabled 0
+	webserver_magic_phrase "I really want an unsecure (non-SSL) Webserver on my coffee machine"
+	webserver_magic_phrase_confirm ""
+	webserver_port 8080
 }
 
 # default de1plus skin
--- a/de1plus/machine.tcl
+++ b/de1plus/machine.tcl
@@ -419,9 +419,10 @@ array set ::settings {
 	scale_stop_at_half_shot 0
 	lock_screen_during_screensaver 0
 	webserver_enabled 0
-	webserver_magic_phrase "I really want an unsecure (non-SSL) Webserver on my coffee machine"
+	webserver_magic_phrase "I really want an unsecured (non-SSL) Webserver on my coffee machine"
 	webserver_magic_phrase_confirm ""
 	webserver_port 8080
+	webserver_authentication_key "ChangeMe"
 }
 
 # default de1plus skin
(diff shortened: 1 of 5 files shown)


miharekar:
Yes, please! 🥳️

Can't wait to configure this with my homebridge so I can say "hey, siri, turn on decent" from my bed when I wake up, and it'll be just warmed up when I reach it 😄️

Mimoja:
What would a proper way of authentication be? Is a simple plain password good enough?

miharekar:
Yeah that would work for me. http basic auth ought to do.
Ideally configurable password with min-length so that to enable the server you would have to enter the password and that already raises the security bar high enough imo.

Damian-AU:
Ah cool, I was going about to google tcl server code this week, see if I could come up with something, this saves me some work.

de1plus/machine.tcl
```diff
@@ -418,6 +418,10 @@ array set ::settings {
 	comms_debugging 0
 	scale_stop_at_half_shot 0
 	lock_screen_during_screensaver 0
+	webserver_enabled 0
+	webserver_magic_phrase "I really want an unsecure (non-SSL) Webserver on my coffee machine"
```
miharekar, line 422: 😂️

Mimoja: I mean... Jeff is not wrong

Mimoja:
Auth added

decentjohn:
I'm really ok with this, and the next step seems to me to be to expose the de1_comms functionality over REST...

I'm not so happy with any default "magic phrase", though.  Can we auto-assign one on app startup, and put it into the settings.tdb, so that only by getting hardware access to settings.tdb, or tablet access to see the magic phrase, could an app do anything with this.

Mimoja:
1) Sure. I just wanted to make sure everyone testing this was aware of the state and implications. Jeff over on Diaspora made a point that it is not necessarily obvious to everyone what it could mean and I think that it is a reasonable approach to be extra careful here. 
2) Do you have an interface in mind that you like for inspiration?

decentjohn:
Well... I prefer power over safety but others might disagree.

Since Tcl commands are always
cmd param1 param2 param3

this naturally maps to a RESTful interface that is:
http://myde1:8080/rest/cmd/param1/param2

thus essentially making a REST->Tcl proxy.

Return results from procs is always a list, so it's natural to return all procs via a JSON list.  

I'll join the API discussion on Basecamp.

Mimoja:
Closed in favour of #41

PR #34, opened 2020-12-12
Fixed slider sync bug

barneyhawes:
Fix for bug spotted by @decentjohn in #32


barneyhawes:
Reimplemented the `variable` property with the correct variable name.

PR #36, opened 2020-12-13
Show android toast on splash screen resize, fixes #35

### LINKED ISSUE #35: When calling photoscale, alert user
On app start up, if the device resolution is new (e.g. because the tablet is new and has a different resolution), the app resizes many images. The only feedback that the user gets is a repeated spinner, and it can seem like the app is frozen, because the spinner repeatedly starts and stops with no other feedback. 

It would be good if there was some sort of text alert—e.g., "Resizing images for your device. Please wait." Or whatever. 

See https://3.basecamp.com/3671212/buckets/7351439/messages/3288128978#__recording_3288326911

Mimoja:

### DIFF
--- a/de1plus/utils.tcl
+++ b/de1plus/utils.tcl
@@ -466,6 +466,7 @@ proc random_saver_file {} {
             set rescale_images_x_ratio [expr {$::screen_size_height / 1600.0}]
             set rescale_images_y_ratio [expr {$::screen_size_width / 2560.0}]
 
+            borg toast [translate "Resizing images"]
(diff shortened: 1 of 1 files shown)


Mimoja:
Confirmed no harm on win32

decentjohn:
I approved it, but then modified it a bit more to give a toast message for each image being converted, as it can take a long time, and also to display the filename of the image being worked on.

And I did this for all images, not just the splashes.

Screenshot_2020-12-13-21-52-29

Mimoja:
Cool, thank you

PR #37, opened 2020-12-15
Add plugin system

Mimoja:
This commit adds a plugin system (without the UI) to be used to
support more extensions to the app that might not be beneficital
to all users. It takes inspiration from the skin subsystem.

Every plugin has its settings (if available) loaded before sourcing.
The settingsfile is hardcoded to $plugin-settings.tdb. The version
translation of settings is within the plugin responsibilities.

Only the top most level of the plugin folder is searched and sourced.
Plugins can therefore store additional data in subfolders.
The use of namespaces in encuraged, but not enforced by the code.

Tested on win32/undroidwish.

This commit also contains an example plugin showing how to trace functions
state-chagnes and work with settings.


Mimoja:
"pluging" /0\

Mimoja:
@barneyhawes can I maybe ask you for feedback?

Mimoja:
Any hints to how to build a UI around this are appreaciated :)
I would like to do in a seperate commit tho

decentjohn:
ps: I'm fine leaving this as "plugins" but the word we use in the UI might be different, ie "extensions"

decentjohn:
Could you please put a catch {} around the loading of each plugin, and if an error occurs, please display the error reported, and remove that plugin from the enabled plugins.

Similarly to what is done with skins.

decentjohn:
Please document how to make your own extension:
https://3.basecamp.com/3671212/buckets/7351439/vaults/3304226704

decentjohn:
I've added a 4th button for the Extensions, and will commit it shortly.
Screen Shot 2020-12-16 at 12 11 35 PM

ebengoechea:
About disabling the plugin if it raises an error on startup, could it be make dependent on a settings option (and ideally skins too)? When developing skins (and plugins now) it's a nightmare to manually re-enable it after each fail. I had to make some dirty hack by Damian to prevent it, having a standard skin/plugin debugging option would be much better.

decentjohn:
Good suggestion. I've done this with skins now:
https://github.com/decentespresso/de1app/commit/897bbc11b4981bfb91ffa06eae11638ab6f447b1

Mimoja:
You can just write "load_plugin myCoolPlugin" anywhere in the startup code. I thought that it was stupid enough to be use it that way :P

de1plus/plugins.tcl
```diff
@@ -0,0 +1,76 @@
...
+}
+
+proc load_plugin {plugin} {
+	if {[catch {
+        create_plugin_namespace $plugin
+        load_plugin_settings $plugin
+		source "[homedir]/[plugin_directory]/$plugin.tcl"
+        ::plugins::${plugin}::main
+	} err] != 0} {
+		catch {
```
Mimoja, line 48: It is catch surrounded already

ebengoechea:
X-D  Easier, yes. I need to better understand everything the app has, Johanna. It's coming slowly.

barneyhawes:
> @barneyhawes can I maybe ask you for feedback?

Sure. Is there anything particular you want feedback on? I'm not sure what the aim are here (did I miss it on Basecamp?) but it would be good to know what this is designed to achieve.

> Any hints to how to build a UI around this are appreaciated :)

Do you mean a UI framework that the plugins can draw on? I created a UI framework for the Metric skin. It is currently specific to the design I created for Metric, but we could do the same for Insight. It would be great to switch over to a code-generated UI instead of the bitmaps.
Or perhaps you mean a UI for selecting plugins?

Mimoja:
@barneyhawes short answer: "yes"
long answer: You write TCL together with @johnbuckman and me here. I value your input and if you dont like something I would love to hear before this makes it into a beta. Yes, I like the Idea of a UI framework, would be amazing to have. Maybe not highest priority right now. With John adding the button to the tab I can work it from there similiar to how the settings are already done, but I would really love to ensure you have everything you want / need for the metric skin available.

barneyhawes:
Thanks for asking for my input.

With the existing skin option being so broad (i.e. allowing much more than just a visual refresh) I can't see that I would use extensions myself, and I can't think of too many reasons for extensions at all. Automatic uploads of shot history is a nice one. Maybe support for another accessory such as a scale.

From a UI point of view I can see pitfalls that could make this painful. It could make it harder for users to find the settings they need. You would need to identify the plug that provides a particular function in order to locate settings associated with it.

I have some ideas for a re-working of the settings screens that I think would be more intuitive and also extensible, but that's probably too big in scope to consider before implementing something like this.

barneyhawes:
I suggest that we allow each plug to specify a display name for the list of plugins (which gets us away from filenames) and a description to be displayed next to the list box when choosing plug ins.

decentjohn:
Main reason for a plugin is to add some functionality that should be available to all skins.

For example "upload my history automatically to the cloud"

Or "generic history viewer"

I've also REALLY wanted to write a "help me dial in my grind" wizard that would step you through several espressos and based on your curve, recommend you change the grind.  That could be a great feature to have on all skins.

decentjohn:
Another... "connect to infrared BLE thermometer and stop steaming automatically when it senses a certain temperature"

https://www.youtube.com/watch?v=VYRXNf_5Oq8

PR #40, opened 2020-12-16
Move plugins into their own folder and build GUI

Mimoja:


decentjohn:
Perhaps using the same color scheme would be good?

Screen Shot 2020-12-17 at 11 27 03 AM

Screen Shot 2020-12-17 at 11 26 51 AM

decentjohn:
Also the dark "selected" bar should likely happen when you tap the selected extension. It does not.

Mimoja:
That is correct, i removed the "make blue" code until I figure out how to render the text next to it properly

Issue #42, opened 2020-12-17
how to configure extensions?


decentjohn:
Maybe if an extension defines a configure proc, you could call it to show controls for it here on this page.

![Screen Shot 2020-12-17 at 11 30 08 AM](https://user-images.githubusercontent.com/11965250/102440607-83100980-405b-11eb-9759-1031976745b2.jpg)

Mimoja:
Yes, i want to do that. The extensions are already exporting their settings in their namespace so that this is explicitly possible.
We also have name / author / description available

Mimoja:
I think this has proven solved :)

PR #43, opened 2020-12-17
Print settings when selecting an extension

Mimoja:
As we do not want to source and extension code without the
uses consent we can not display any dynamic variables.
But: We can use the additional space to show the saved settings
that are in use. This allows us to double-check the current configuration
before activating a plugin. We can not easily open the settings file
in an external android editor without the risk of exposing user data to
3rd party apps without the user consent. I personally dont think we can
solve that from within the app (for now). User will still need to edit their
plugin settings from within the filebrowser / adc / their app of choice.

### DIFF
--- a/de1plus/plugins/example/plugin.tcl
+++ b/de1plus/plugins/example/plugin.tcl
@@ -1,21 +1,52 @@
+
+# Change package name for you extension / plugin
 set plugin_name "example"
 
+# These are chown in the plugin selection page
 set ::plugins::${plugin_name}::author "JoJo"
 set ::plugins::${plugin_name}::contact "email@coffee-mail.de"
 set ::plugins::${plugin_name}::version 1.0
-set ::plugins::${plugin_name}::description "Minimal plugin to showcase the interface"
+set ::plugins::${plugin_name}::description "Minimal plugin to showcase the interface of the plugin / extensions system."
+
+# preload is called on app start even if the plugin is disabled. Can be used to show
+# dynamic informations on the settings overview. Please dont put logic here
+# needs to return the page you want to be shown first
+proc ::plugins::${plugin_name}::preload {} {
+    # Unique name per page
+    set page_name "plugin_example_page_default"
+
+    # Background image and "Done" button
+    add_de1_page "$page_name" "settings_message.png" "default"
+    add_de1_text $page_name 1280 1310 -text [translate "Done"] -font Helv_10_bold -fill "#fAfBff" -anchor "center"
+	add_de1_button $page_name {say [translate {Done}] $::settings(sound_button_in); fill_extensions_listbox; page_to_show_when_off extensions; set_extensions_scrollbar_dimensions}  980 1210 1580 1410 ""
+
+    # Headline
+    add_de1_text $page_name 1280 300 -text [translate "Example Plugin"] -font Helv_20_bold -width 1200 -fill "#444444" -anchor "center" -justify "center"
+
+    # The actual content. Here a list of all settings for this plugin
+    set content_textfield [add_de1_text $page_name 600 380 -text  "" -font global_font -width 600 -fill "#444444" -anchor "nw" -justify "left" ]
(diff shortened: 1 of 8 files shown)


Mimoja:
grafik

Mimoja:
As plugins were explicitly needed to provide the most dangerous functionalities (aka IoT things) we should not live-reload any settings here as they can lead to major security issues (and just restarting the app after a config change does not hurt _that_ much). We cannot guarantee that each and every plugin handles the change of config gracefully at runtime and I doubt we want to open that can of worms right now.

Mimoja:
fixes #42 (-ish)

barneyhawes:
Can I suggest that now is the time to decide that plug ins /should/ be able to adapt to changing settings on the fly?

Or better still, settings should have attributes such as "requires reload". This would also be a way to specify how settings are edited - eg textbox, value slider, checkbox.

Mimoja:
Imagine the case that a changed variable changes the apps startup behaviour. How - without restarting the app - could that be implemented? Lets say: Adding a scale via bluetooth in contrast to ble. Requireing the app to never have connected to the same device via ble.

decentjohn:
My suggestion is that until the [x] is clicked, no loading of the module happens.

Once clicked, we can load the module and call it's "settings()" function (or whatever you call it) which should (if applicable) have some code that loads controls onto this page, like so:

`add_de1_widget "extension_config" checkbutton 1538 830 {} -text [translate "your national ID number, please..."] -padx 0 -pady 0 -indicatoron true  -font Helv_9_bold -anchor nw -foreground #7f879a -activeforeground #7f879a -variable ::current_adv_step(exit_if)  -borderwidth 0  -highlightthickness 0  -command save_current_adv_shot_step -selectcolor #f9f9f9 -activebackground #f9f9f9 -bg #f9f9f9 -relief flat `

If we're quite nervous about extensions doing bad things, we could wrap them inside a safetcl interpreter. https://www.tcl.tk/software/plugin/safetcl.html

However, I don't think we're going to solve any security problems easily, that can't be bypassed.

My preference would be to only include vetted extensions with the de1app, from people known on the Diaspora forum. And all source is open, so it'll be not-so-easy to make something 'evil' that survives for long.

Note that extensions will also be going through the 'beta' app cycle, and not included for quite a while.

Mimoja:
grafik

@decentjohn if i create a new page, page_to_show_when_off it and then do the same thing going back i get overdraws. Any Idea how to deal with it?

de1plus/plugins/example/plugin.tcl
```diff
@@ -1,21 +1,52 @@
...
+# only set if you want to provide a settings page for your plugin
+# returns the name of the page to show next
+proc ::plugins::${plugin_name}::settingsfunction {} {
+    # Unique name per page
+    set page_name "plugin_example_page_default"
+
+    # Background image and "Done" button
+    add_de1_page "$page_name" "settings_message.png" "default"
+    add_de1_text $page_name 1280 1310 -text [translate "Done"] -font Helv_10_bold -fill "#fAfBff" -anchor "center"
+	add_de1_button $page_name {say [translate {Done}] $::settings(sound_button_in); fill_extensions_listbox; page_to_show_when_off extensions; set_extensions_scrollbar_dimensions}  980 1210 1580 1410 ""
```
Mimoja, line 20: Do I need to something special to go back?

decentjohn: Are you dynamically creating new pages?  I've not done that, I create them all when the app launches.

If not, then I recommend you look at the log (make sure it's being created) as page_to_show_when_off will automatically hide elements that are not part of that context and you can confirm that's the case.  Make sure in the log that page to show / page to hide, is what you expect.

There's not enough code above for me to guess what the issue is, and using $page_name is hiding the context info from me, so I can't really review it.

Mimoja:
fixed the rendering issue, should be fine for testing now

PR #45, opened 2020-12-21
Scale disconnect

Mimoja:
Not tested on windows

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -105,7 +105,19 @@ proc scale_enable_grams {} {
 	}
 }
 
+proc scale_check_alive {} {
+	set since_last_ping [expr {[clock seconds] - $::de1(last_ping)}]
+
+	if {$since_last_ping > 5} {
+		set $::de1(scale_device_handle) 0
+	}
+}
+
 proc handle_new_weight_from_scale { sensorweight scale_refresh_rate } {
+
+	set ::de1(last_scale_ping) [clock seconds]
+
+	after 5000 scale_check_alive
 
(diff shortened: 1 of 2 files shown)


decentjohn:
I believe this implementation suffers from several bugs:

1- 10x per second (for each weight received) a scheduled proc will run 5 seconds later.  That means 50 scheduled procs will be waiting, at all times.  I don't see the point of this.

2- you're removing the BLE handle but not disconnecting. If the Bluetooth recovers automatically, then we'll have lost the handle and won't be notified of either a recovery, or disconnection, and thus have permanently lost our ability to talk to a scale.

What I do with the DE1 timeout mechanism:
1- keep track of last ping without a scheduled proc
2- set a separate boolean that indicates no weight being received, but do not actually mess with the BLE connection, as it might recover on its own
3- the gui displays "disconnected" when no update has been received, but no other action is taken
4-if we do this same thing with the scale, then WEIGHT will blink after 5 seconds of no data, and the user can choose to tap that label to initiate a scale disconnect/reconnect.

Mimoja:
1) yes , i see how that is wasteful
2) i don't think BLE does that. I tested it with an esp32 that just delayed the (fake) weight data by 6 seconds once and then sends normal. I did not have any Problems reconnecting through the app in that case. Maybe we don't need to care for the potential reconnecting by recovery case ? 
(Will read spec and do additional testing first)

decentjohn:
I think I'd prefer that you check the last ping time in
proc waterweight_label_text{}

and cause the blink there.

I can do this in 5 minutes' work, if you're unsure what I want.

PR #46, opened 2020-12-30
Translation support

Mimoja:
Not fully tested, please do not merge. Build on top of the settings screen. If that is fixed, this should just fall in place (fingers crossed)


Mimoja:
As discussed on basecamp

PR #48, opened 2021-01-15
Fixing Visualizer plugin for proper upload.

JakubOleksy:
I fixed this plugin to properly follow the multipart/form-data protocol. Tested locally and confirmed it works.

### DIFF
--- a/de1plus/plugins/visualizer_upload/plugin.tcl
+++ b/de1plus/plugins/visualizer_upload/plugin.tcl
@@ -1,6 +1,7 @@
 package require http
 package require tls
 
+
 set plugin_name "visualizer_upload"
 
 set ::plugins::${plugin_name}::author "Johanna Schander"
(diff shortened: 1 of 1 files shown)


JakubOleksy:
Flagging @Mimoja and @decentjohn

Mimoja:
Heyya :)
The plugin feature is not yet finished and therefore no development is happening to the plugin POCs. Please be patient for the moment

Mimoja:
Good catch tho. Would never have the ; expected to be the issue

JakubOleksy:
It's much more than the ';'. You need to change the structure of the message to comply with the multipart format.

de1plus/plugins/visualizer_upload/plugin.tcl
```diff
@@ -17,16 +18,17 @@ proc upload {content} {
     set password $::plugins::visualizer_upload::settings(visualizer_password)
 
     set auth "Basic [binary encode base64 $username:$password]"
-    # For the moment we need to butcher the form-data format
-    set type "multipart/form-data; charset=utf-8;"
-
-    set headerl [list Authorization "$auth" Content-Type "$type"]
+    set boundary "--------[clock seconds]"
+    set type "multipart/form-data, boundary=$boundary"
```
miharekar, line 22: I think you should add `charset=utf-8` back

Mimoja: okay, will do.

At some point we had the exact same structure, i changed it after Miha told me to remove the boundaries and just send data. Which worked for small files (forwhatever dark magic reason).

Mimoja:
okay, will merge this after checking the utf-8 param.

At some point we had the exact same structure, i changed it after @miharekar  told me to remove the boundaries and just send data. Which worked for small files (forwhatever dark magic reason).

miharekar:
The app expects UTF-8 so with a degree symbol in profile title for example I'm getting
`invalid byte sequence for encoding "UTF8": 0xb0` 
errors on server.

Yeah I never built boundaries by hand in any of the languages I've ever worked with so I was assuming there had to be a better way 🙈

Mimoja:
@miharekar can you confirm that the last upload was fine?

miharekar:
@Mimoja this is the last one on your user: https://visualizer.coffee/shots/4a31152a-eb03-45f9-87e1-dc99c833546d

It looks very short, but maybe that's all there is? 🤷‍♂️😅

PR #50, opened 2021-01-16
plugins: Add plugin tabs

Mimoja:
In this commit we are removing the scentone heart button from the Insight skin
(so far without replacement) and replace it with a plugin Tab overview.
Plugins can be attached to Tabs if they support this. To support this feature
they need to support a preload method that hand them the context to draw into.
Therefore an App restart is required between Tab changes.

We are also updating the psd submodule at once to include the new tab base files.


Mimoja:
grafik
grafik

Mimoja:
The PSD pr de1plus_psd#2 needs to be merged first for the submodule commit to be available.
(Just click rebase and merge)

PR #52, opened 2021-01-27
Fix physical keyboard command functionality. Update comments for proc…

vjapolitzer:
This is a simple fix for physical keyboard command functionality. Additionally, I have updated the comments in proc handle_keypress.

### DIFF
--- a/de1plus/gui.tcl
+++ b/de1plus/gui.tcl
@@ -1512,9 +1512,10 @@ proc page_display_change {page_to_hide page_to_show} {
 }
 
 proc hide_android_keyboard {} {
-	# make sure on-screen keyboard doesn't auto-pop up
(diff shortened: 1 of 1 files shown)


Mimoja (approved):
Looks good to me. It was confirmed on the forums that this  was the fix required

vjapolitzer:
Ah, I just got my Android 9.0 tablet today, and unfortunately discovered that the CTRL modifier hack no longer works for the idle command with `space`, so in the new commit I propose changing from `space` to `i`, which works on my old Android 7.1 tablet, as well as 9.0.

vjapolitzer:
(Silly goof, it's Android 7.0, not 7.1. Force pushed from my work computer and forgot I had it configured to my other git account and didn't want it as a contributor... SMH. All good now.)

Issue #29, opened 2020-12-02
include step transitions in the history file


decentjohn:

Mimoja:
addressed with #54, following fix #53

PR #54, opened 2021-01-31
Shot file improvements

Mimoja:
Add additional variables to espresso log files

As we are no longer just using the App to save, edit and view the shot files
it is time to add the missing variables to shot files.
This includes in this commit:
- step transitions
- resistance
- deltas
As the App might change the way it is calculating things in the future it
is unrealistic for 3rd party devs to reimplement and keep up with all changes.
To easy traceability a app_version field was added on the highest shot file level.

PLEASE MERGE #53 first


Mimoja:
Fixes #29

ebengoechea:
Great, was looking forward to this!
What about the firmware version?

PR #56, opened 2021-02-02
Improving Mass and Mass-Flow Estimates

jeffsf:
Commit messages should be self-explanatory. Various Basecamp threads for context.


de1plus/updater.tcl
```diff
@@ -133,7 +135,7 @@ proc percent20encode {in} {
     set out $in
     regsub -all " " $out "%20" out
     #regsub -all "&" $out "%26" out
-    regsub -all {"} $out "%22" out
+    regsub -all "\"" $out "%22" out
```
Mimoja, line 138: Why was this change necessary?

Mimoja (approved):
I appreciate the cleanup, looks good to me.

Mimoja (approved):
Yassss!!

Mimoja (commented):
Was this tested on Android 5 with the "old" BLE stack?

de1plus/de1_de1.tcl
```diff
@@ -0,0 +1,221 @@
...
+#
+# These are not in machine.tcl due to apparent assumptions on inclusion order
+#
+
+package provide de1_de1 1.0
+
+package require de1_event 1.0
+package require de1_logging 1.0
+
+namespace eval ::de1::state {
```
Mimoja, line 14: We need to make the descision if @johnbuckman @decentjohn s statement:
"Keep it as simple as possible, no namespaces, no lambdas, nothing magical"
is still worth keeping. While it might be nice for new devs, namespaces are imho still trivial enough to not overcomplicate our lives.

Mimoja:
I find this code almost impossible to review as so many things are changing in the complete oposite to how they used to be. I appreciate the cleanup efforts a lot. I have not made up my mind in regards of the event based system, but dont really hate it. It just seems a bit overengeneered and a bit to C++-ish to me.
Will test this on my machine in the next days and report back if I find something breaking.
Thank you a lot for tackeling this!

de1plus/updater.tcl
jeffsf: It causes syntax validation problems in at least Tcl mode of emacs

No functional change, just a representational change of a string consisting of a the single "double quote" character

jeffsf:
I only have the tablet that shipped with the v1.3 which is, as I recall Android 8.1

The BLE module comes from Androwish and the basic interactions with that module in `bluetooth.tcl` are not changed. A diagnostic to reveal the hex of the code if not recognized was added. The majority of changes related to scale communication, itself, included reducing the number of tare requests and bringing them together (previously they were in `binary.tcl` and `machine.tcl`, as well as in `bluetooth.tcl`), as well as repeating the request for updates, which seems to get "lost" in startup situations, resulting in a "connected" scale that isn't sending updates. In some cases the scale was getting bombarded by tare requests, from multiple places, and generating new ones on every weight packet received (Skale 2 clocks packets somewhat strangely, often resulting in two weight packets in tens of milliseconds).  It wouldn't surprise me if similar "overload" problems are behind the recently reported Skale shows connected, but no display symptoms, as it appears that the "APP" connection is followed by a hardware display off and waiting for the app to turn on the display.

I agree that the event system being extensible isn't needed for "just" the scale, but I was considering that things like the catering/refill kits would need to be dealing with water-low notifications, as well as new notifications coming from the "ultra-advanced" profiles. I provided the generalized forms to be able to isolate the implementation details of registering and calling listeners from what could be a growing number of event families. I did not implement query, delete, replace, ... operators as I didn't see a need for them at this time.

Please let me know if you have any issues with your testing, especially with non-Skale units.

If you have some data that shows the severe kind of scale excursions John was reporting around users of rinsers and knock boxes (30 g???), I'd appreciate a copy of the `.shot` file so that I can better tune the median filters, or determine that a different approach would be needed. (The existing `.shot` data, unfortunately, is both heavily filtered as well as on the DE1 25/100 or 25/120 second time base.)

jeffsf:
I'll merge the contents of 590bb70 next time I need to rebase due to upstream changes, or as comments are addressed.

The added commit may address some of the issues being reported with drink_weight showing higher than scale weight. One recent example is  _Shot Recording (final weight)_ from February 1st.

Having full-resolution, unfiltered scale data will help diagnose these issues in the future.

As a note, my current SAW lead time is 0.2 seconds with these changes.

de1plus/de1_de1.tcl
jeffsf: The couple of places I used lambda expressions were to keep the existing method signatures the same. At least with the `lambda` package they end up looking like a "nameless" proc.

```
	[lambda {event_dict} {
		save_this_espresso_to_history \
			[dict get $event_dict previous_state] \
			[dict get $event_dict this_state] \
		} ]
````
reads very similarly to

```
	proc save_this_espresso_to_history_for_event_callback {event_dict} {
		save_this_espresso_to_history \
			[dict get $event_dict previous_state] \
			[dict get $event_dict this_state] \
		} 
```
without leading one to wonder, "Which of these two nearly identically named functions should I be using?" or why the `{} {}` construct exists in 
```
proc save_espresso_rating_to_history {} {
	#unset -nocomplain ::settings(history_saved)
	set ::settings(history_saved) 0
	save_this_espresso_to_history {} {}
}
```

---

At least for me, keeping the code organized by subsystem or function, is helpful. For me, it improves readability, as well as reducing the risk of one bit of code walking over another's globals.

**Comment by jeffsf** [comment]
Updated to include previously discussed "fixup" commits. Merge base continues to be tag v1.34.5 so early commits around logging and `.gitignore` should be unchanged.

Also:

Adds `::de1::state::current_state` and `current_substate` to localize coupling to the internal state variables and their number-to-name translation arrays as well as providing extension writers a straightforward way to readably test the DE1 states.

Moved table of scale lag to `::device::scale::sensor_lag` as it is a property of the scale, not any of the estimation algorithms

Removed internal tracking of flow estimates during the shot as they were unused. Originally intended for resampling the data on the scale's sample clock to those of the other graphed data on the DE1 sample clock. As `blt::graph` can draw the series on the graphs without resampling or rescaling, this is unnecessary to achieve the goal.

Removed debug use of the call to start/stop the atomaxskale timer to reduce unnecessary BLE traffic. The Skale 2 only has a single display which is used to display weight and the scale's timer is never retrieved by the app. It can easily be overriden by setting `run_timer` on scale initialization.

Added `on_update_available` event to the scale, called at the completion of `process_weight_update`. This permits easy extension of the graphs or other processes related to weight/flow in an efficient manner. As one example, here's how I'm looking at high-resolution, mass-flow data. This code also shows how the predicates allow one to write extensions without needing to try to trace down "magic numbers" in the code.

```
namespace eval ::device::scale::hack {

	proc init {} {

		blt::vector create ::device::scale::hack::flow_rate
		blt::vector create ::device::scale::hack::flow_time
	}



	proc on_major_state_change {event_dict} {

		if { [::de1::state::is_flow_state [dict get $event_dict this_state]] } {

			::device::scale::hack::flow_rate length 0
			::device::scale::hack::flow_time length 0
		 }
	}

	proc on_update_available {event_dict} {

		if { [::device::scale::history::is_recording] && [::de1::state::is_flow_during_state] } {

			::device::scale::hack::flow_rate append [::device::scale::history::flow_estimate_raw]

			# Time estimate already includes algorithm compensation
			::device::scale::hack::flow_time append \
				[expr { [::device::scale::history::flow_estimate_raw_time] \
						- [expr { $::timers(espresso_start) / 1000.0 }] \
						- [::device::scale::sensor_lag] }]
		}
	}
}

::device::scale::hack::init

::de1::event::listener::on_major_state_change_add ::device::scale::hack::on_major_state_change
::device::scale::event::listener::on_update_available_add ::device::scale::hack::on_update_available
```

**Comment by jeffsf** [comment]
Resetting the estimators should have no impact in normal operations, but there was the possibility of some strange beginning-of-shot effects with either the former, exponential-average estimators or even the one-second estimators. If, for example, someone added or removed a heavy mug just before the flow started on the shot, the scale would tare. This would appear perviously (including in the current Stable and Beta code) to be a significant change in weight. It would persist in the flow estimate for several seconds with the former, exponential-average estimators (5 tau ~ 10 seconds). Though this would likely be during "preinfusion", it would still produce phantom estimates.

As `warn_if_scale_not_reporting` was already gated by checking for an SAW target, the redundant check in the proc itself was removed. This permits it to be used any time a process or user should be notified that updates are not being received, whether SAW is active at that moment or not.

**Comment by jeffsf** [comment]
Included in #79 which includes #59

Rebased on
```
$ git log -1 $(git merge-base main HEAD)
commit 2ab1d98 (upstream/main, upstream/HEAD, mini/main, main)
Author: Mimoja <git@mimoja.de>
Date:   Wed Feb 24 11:17:57 2021 +0100

    example_plugin: use local settings variable to clean up logic
    
```

jeffsf:
Closing in favor of adoption of #79, which contains these commits

PR #59, opened 2021-02-14
Correct errors in volume estimation and line frequency

jeffsf:
This PR builds on #56 as it requires the logging and event systems implemented there.

The two new commits are 

https://github.com/decentespresso/de1app/pull/59/commits/9866b2d71201f81a273d27ab139950f24ecc72bf
https://github.com/decentespresso/de1app/pull/59/commits/d9335a2568a1b9104d091a1748e796833973c690

In working toward being able to time-align DE1 and scale data, errors in the use of the DE1 internal half-cycle counter were discovered.

These errors result in the use of a 1/100ths of a second time base for all users, though users on 60 Hz have a 1/120th half-cycle time. This led to `::espresso_water_dispensed` being 20% off for 60 Hz users. When last discussed, Ray indicated that the DE1 flow data should be dimensionally correct in ml/s, no matter if 50 or 60 Hz. This, however, is an app-side issue.

The 16-bit counter from the DE1 was also not being unwrapped, leading to periodic errors in the computation of water dispensed.


jeffsf:
Removed stale comment

jeffsf:
Added user and developer notes to commit message.

Corrected some typos and restored logic around trying to restart an already running estimation.

Tested DE+ v1.3, 120 V, 60 Hz, stock tablet, Skale 2

decentjohn:
Ah, I see the problem.  Hertz is hardcoded in my app to 50:
hertz 50

probably because we (long ago) didn't report hertz over BLE. 

I'm asking if I can get a hertz value from the firmware from Ray,  That should fix his issue, I think, no?

There is code in there to deal with this already, but it's not aware of your real hertz.
```
  	set spec {
		Timer {Short {} {} {unsigned} {int(100 * ($val / ($::de1(hertz) * 2.0)))}}

```

One way you could get this @jeffsf is to change the hertz value in machine.tcl and see if that does the job.

jeffsf:
For a variety of reasons, including being able to reconstruct the DE1's time base, keeping the reported, 16-bit value is important. On the v1.3 stock tablet, the packet-arrival time over BLE for the DE1 has a standard deviation of around 40 ms, and delays of up to 200 ms are not unusual. The existing conversion from the raw data suffers from a loss of precision as well. While 50 Hz is conveniently 250 ms nominal between reports, at 60 Hz 25 zero-crossings is 25/120 ~ 20.833333... 100ths. 

At least in the code as it exists on Stable and Beta, this is only used for the calculation of flow accumulation. However, it is important to preserve in its 16-bit form to be able to properly unwrap it (corrected in this PR) and to be able to deduce the likely DE1 sample time to be able to time-align it with scale data.

I've already validated that this is a problem and can provide hours of packet-arrival data to support the change. 

If the line frequency (and voltage) is available directly from the DE1, as noted in the commit messages, using that would be preferred for its simplicity. 

In addition to the fixed-Hz problem, this PR fixes the failure to unwrap the counter, maintains access to the counter, and introduces an event on which GUI/skin implementations can be informed of changes. Even with the extension system, to make any change, including a simple addition, the entirety of one or more proc definitions need to be pulled into the skin or extension. SWDark4 already has two, obsolete copies of `append_live_data_to_espresso_chart` (see #64) and DSx replicates it as well. This becomes a never-ending maintenance issue requiring a rebase of skin changes on a different file entirely. Between performance considerations with a real-time, though single-threaded application and the variables within scope of a proc not being available to `trace` (except for the line-by-line versions), use of Tcl `trace` is not a reasonable alternative.

jeffsf:
Included in #79, builds on #56

Rebased on

$ git log -1 $(git merge-base main HEAD)
commit 2ab1d98 (upstream/main, upstream/HEAD, mini/main, main)
Author: Mimoja <git@mimoja.de>
Date:   Wed Feb 24 11:17:57 2021 +0100

    example_plugin: use local settings variable to clean up logic
    
jeffsf:
Closing in favor of #79, which contains these commits

PR #60, opened 2021-02-17
Uses new global var 'loaded_plugins' to avoid reloading a plugin

ebengoechea:
Change to avoid reloading plugins, I have used a new global var 'loaded_plugins', maybe you prefer a variable in the plugins namespace?

### DIFF
--- a/de1plus/plugins.tcl
+++ b/de1plus/plugins.tcl
@@ -61,8 +61,13 @@ proc plugin_preload {plugin} {
 }
 
 proc load_plugin {plugin} {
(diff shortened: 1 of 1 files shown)


Mimoja (approved):

jeffsf:
Personal opinion is that the `::plugins` namespace is a better choice.

ebengoechea:
Agree.

ebengoechea:
This PR should be cancelled, as it has been superseded by PR #63

PR #65, opened 2021-02-20
machine: Complete state and substate arrays

jeffsf:
After noting that substate 14 was returned with FW 1238 and that `::de1_substate_types()` did not include it, complete the state and substate lists to be consistent with APIDataTypes.hpp

Based on APIDataTypes.hpp

### DIFF
--- a/de1plus/machine.tcl
+++ b/de1plus/machine.tcl
@@ -488,6 +488,7 @@ array set ::de1_num_state {
   18 Clean
   19 InBootLoader
   20 AirPurge
+  21 SchedIdle
 }
 
 
@@ -505,7 +506,18 @@ array set ::de1_substate_types {
 	4	"preinfusion"
 	5	"pouring"
 	6	"ending"
+	7	"Steaming"
+	8	"DescaleInit"
+	9	"DescaleFillGroup"
+	10	"DescaleReturn"
+	11	"DescaleGroup"
+	12	"DescaleSteam"
+	13	"CleanInit"
+	14	"CleanFillGroup"
+	15	"CleanSoak"
+	16	"CleanGroup"
 	17  "refill"
+	18	"PausedSteam"
 }
 array set ::de1_substate_types_reversed [reverse_array ::de1_substate_types]
 


de1plus/machine.tcl
```diff
@@ -505,7 +506,18 @@ array set ::de1_substate_types {
 	4	"preinfusion"
 	5	"pouring"
 	6	"ending"
+	7	"Steaming"
```
Mimoja, line 509: I think CamelCase vs all-lowercase got mixed up here.

Mimoja (changes requested):

de1plus/machine.tcl
jeffsf: What would you prefer? 

The names I inserted are right out of the struct. If the API is the struct, then those are correct. See `APIDataTypes.hpp`

Regrettably, changing the others at this point is a major refactor, as they have been used extensively as keys throughout the current app.

```
  Sleep,      // 0 Everything is off
  GoingToSleep, // 1
  Idle,       // 2 Heaters are controlled, tank water will be heated if required.
  Busy,       // 3 Firmware is doing something you can't interrupt (eg. cooling down water heater after a shot, calibrating sensors on startup).
  Espresso,   // 4 Making espresso
  Steam,      // 5 Making steam
  HotWater,   // 6 Making hot water
  ShortCal,   // 7 Running a short calibration
  SelfTest,   // 8 Checking as much as possible within the firmware. Probably only used during manufacture or repair.
  LongCal,    // 9 Long and involved calibration, possibly involving user interaction. (See substates below, for cases like that).
  Descale,    // A Descale the whole bang-tooty
  FatalError, // B Something has gone horribly wrong
  Init,       // C Machine has not been run yet
  NoRequest,  // D State for T_RequestedState. Means nothing is specifically requested
  SkipToNext, // E In Espresso, skip to next frame. Others, go to Idle if possible
  HotWaterRinse, // F Produce hot water at whatever temperature is available
  SteamRinse,    // 10 Produce a blast of steam
  Refill,        // 11 Attempting, or needs, a refill.
  Clean,         // 12 Clean group head
  InBootLoader,  // 13 The main firmware has not run for some reason. Bootloader is active.
  AirPurge,      // 14 Air purge.
  SchedIdle      // 15 Scheduled wake up idle state
```

jeffsf:
What does John want here? 

If the API is the C struct, which it arguably is, then the lower-case ones are technically incorrect. 

The existing keys are, however, used extensively in the current code as textual keys, so changing them is a significant refactor with little benefit. Thankfully I did not find them rendered directly in the GUI, with or without `translate` invoked.

```
enum T_Enum_API_Substates : U8 {
  NoState,          // 0 State is not relevant
  HeatWaterTank,    // 1 Cold water is not hot enough. Heating hot water tank.
  HeatWaterHeater,  // 2 Warm up hot water heater for shot.
  StabilizeMixTemp, // 3 Stabilize mix temp and get entire water path up to temperature.
  PreInfuse,        // 4 Espresso only. Hot Water and Steam will skip this state.
  Pour,             // 5 Not used in Steam
  Flush,            // 6 Espresso only, atm
  Steaming,         // 7 Steam only
  DescaleInit,      // Starting descale
  DescaleFillGroup, // get some descaling solution into the group and let it sit
  DescaleReturn,    // descaling internals
  DescaleGroup,     // descaling group
  DescaleSteam,     // descaling steam
  CleanInit,        // Starting clean
  CleanFillGroup,   // Fill the group
  CleanSoak,        // Wait for 60 seconds so we soak the group head
  CleanGroup,       // Flush through group
  PausedRefill,     // Have we given up on a refill
  PausedSteam,      // Are we paused in steam?
```

Mimoja:
I will take the time and go hunt all the usages of this so that we can properly CamelCase it

jeffsf:
I took a cut at finding and replacing the "old" substates, tuning a `sed` script to do most of the work.

The results, the script and associated files are at https://github.com/jeffsf/de1app/commits/substate-replacements

This is not a PR, but hopefully can help to find some of the peskier appearances.

As the repo doesn't contain the submodules, they aren't included.

```
de1app/de1plus]$ git submodule foreach git diff --ignore-space-at-eol
Entering 'plugins/DGUI'
Entering 'plugins/SDB'
Entering 'plugins/keyboard_control'
Entering 'skins/MimojaCafe'
diff --git a/skin.tcl b/skin.tcl
index b81282e..bdc1754 100644
--- a/skin.tcl
+++ b/skin.tcl
@@ -235,7 +235,7 @@ set ::iconik_min_flow 20
 
 proc iconik_get_max_pressure {} {
        if {$::de1_num_state($::de1(state)) == "Espresso"} {
-               if {$::de1(substate) >= $::de1_substate_types_reversed(pouring)} {
+               if {$::de1(substate) >= $::de1_substate_types_reversed(Pour)} {
                        if {$::de1(pressure) >= $::iconik_max_pressure} {
                                set ::iconik_max_pressure $::de1(pressure)
                        }
@@ -248,7 +248,7 @@ proc iconik_get_max_pressure {} {
 
 proc iconik_get_min_flow {} {
        if {$::de1_num_state($::de1(state)) == "Espresso"} {
-               if {$::de1_substate_types($::de1(substate)) == "pouring"} {
+               if {$::de1_substate_types($::de1(substate)) == "Pour"} {
                        if {$::de1(flow) <= $::iconik_min_flow} {
                                set ::iconik_min_flow $::de1(flow)
                        }
Entering '../misc'
Entering '../psd'
```

decentjohn:
Hi Jeff, has this been tested on your end?  If so, I'm ok pushing it to BETA.

jeffsf:
Switching to Draft as, if done "properly" as @Mimoja suggests 9and I agree), this is a very invasive change, potentially breaking (missed substate changes or detection, for example).

I'd definitely prefer to work together on this to get consistent names here, as well as include the error statuses from the C API. 

Because of the risk involved with the change of existing names, this really should be done on staging trees and examined by several of us before moving it to a public repo. I suspect that we'll all find at least one that we've missed. Getting it into a clean, single commit would be my preference.

decentjohn:
Gotcha, thanks for the comments.

jeffsf:
Impacts users running cleaning cycles.

Let's get it in for this release as-is and can address consistency and the error states in the next release cycle.

PR #68, opened 2021-02-22
Add web_api handler to return the DE1 machine state

mariusstrom:
Add checkStatus procedure in web_api extension. I went with a plaint/text return instead of application/json as I just wanted to return a simple 1 or 0 which is what homebridge-http-switch expects.

This is my first time touching TCL so won't be upset if I get yelled at for failing to follow some basic norms. :)

### DIFF
--- a/de1plus/plugins/web_api/plugin.tcl
+++ b/de1plus/plugins/web_api/plugin.tcl
@@ -5,8 +5,8 @@ set plugin_name "web_api"
 
 set ::plugins::${plugin_name}::author "Johanna Schander"
 set ::plugins::${plugin_name}::contact "coffee-plugins@mimoja.de"
-set ::plugins::${plugin_name}::version 1.0
-set ::plugins::${plugin_name}::description "Minimal webserver to toggle the DE1s power state"
+set ::plugins::${plugin_name}::version 1.1
+set ::plugins::${plugin_name}::description "Minimal webserver to report, enable, and disable the DE1's power state"
 
 proc ::plugins::${plugin_name}::main {} {
 
@@ -22,6 +22,7 @@ proc ::plugins::${plugin_name}::main {} {
     # Define handlers
     ::wibble::handle /on togglePowerOn
     ::wibble::handle /off togglePowerOff
+    ::wibble::handle /status checkStatus
     ::wibble::handle / notfound
 
     # Start a server and enter the event loop if not already there.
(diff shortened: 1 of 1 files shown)


de1plus/plugins/web_api/plugin.tcl
```diff
@@ -77,5 +78,25 @@ proc ::wibble::togglePowerOff {state} {
...
+
+proc ::wibble::checkStatus {state} {
+    if { ![check_auth $state] } {
+        return;
+    }
+
+    dict set response status 200
+    dict set state response header content-type "" {text/plain charset utf-8}
+    
+    # Returning a simple text 1 if the machine is in anything other than an idle state. Return text 0 if idle.
```
jeffsf, line 92: Might want to update the comment here to reflect the logic below

de1plus/plugins/web_api/plugin.tcl
```diff
@@ -77,5 +78,25 @@ proc ::wibble::togglePowerOff {state} {
...
+    # Return values chosen by cribbing from Supereg/homebridge-http-switch
+
+    if { $::de1_num_state($::de1(state)) != "Sleep" } {
```
jeffsf, line 95: Looking at the states in `machine.tcl`, you might want to decide which states are "status good", or just return the state name in text and let the caller decide which are "good" for what they want to do next. I don't know that you can, for example, interrupt `Descale` and some others.

If you do decide to digest several states into a Boolean, I've found the Tcl [`in`/`ni`](https://www.tcl.tk/man/tcl8.6/TclCmd/expr.htm#M15) operator useful.

mariusstrom: Hi jeff - nothing here interrupts, it simply returns whether the machine is sleeping or not. It's primarily intended for use so that a smarthome hub can know whether the device is on or off in the event it was controlled independently from the hub.

de1plus/plugins/web_api/plugin.tcl
mariusstrom: Made a subsequent update to change the comment to refer to sleep states rather than idle states, good catch.

Mimoja:
Thanks @jeffsf for the review. Looks fine to me. Will not merge it for now, as I dont have the time to test and verify, no promises to when that is the case, hopefully later today

miharekar:
Just tested it out. Works great.

Maybe we don't even need to check auth for this?

mariusstrom:
> Maybe we don't even need to check auth for this?

@miharekar I considered that, but in the absence of definitive guidance and a threat model decided to leave the auth check on. Agree with you though: this is a safe read-only and unprivileged operation so I would agree that auth may not be required.

Curious if @Mimoja has any thoughts. I think the de1app maturity is pretty in its current architecture is far away from a consolidated threat model, so I feel like it could be better to be too conservative on this front for now.

Mimoja:
Please keep the auth. Things connected to the internet (which will happen one way or another) should never offer metadata for free

Mimoja:
merged

PR #69, opened 2021-02-22
Remove overwrite of the wrong widget 'widget_profile_name_to_save'

ebengoechea:
The removed line of code was overwritting the wrong global variable. It was reassigning the rename-profile widget stored in "widget_profile_name_to_save" to the username widget in the plugin. This means that code in the base app that used that global variable (like proc profile_has_changed_set_colors) was doing the changes to the wrong widget if this plugin was enabled.

Beware that this line of code seems to have slipped by "copy-paste" to other places outside of the main repository, too. I could find instances of it in the MimojaCafe skin and in the keyboard_control plugin. I don't have those repos cloned, so may I please ask @mimoja and @vjapolitzer to look for that line and remove it, or assign $widget to your own widget variable (though that doesn't seem needed as if you had needed the widget reference you would have likely caught the bug).


vjapolitzer:
Thanks for the catch, @ebengoechea. This line is not needed in keyboard_control, so I have removed it and pushed the update.

Mimoja:
merged

PR #74, opened 2021-02-23
Visualizer upload adopt dye changes

ebengoechea:
This PR incorporates the changes I had made in the DYE plugin to the visualizer_upload code. So, in the forthcoming version of DYE as a de1app plugin it can directly use @Mimoja's plugin.

- New items in the settings array to capture the last upload shot clock, result and ID. The result of this last upload is now shown on the settings page, so users can see if their last shot was uploaded correctly or not, and if not the exact reason why, which should reduce users questions and facilitate solving problems. If the shot succeeded, a "text button" is shown to open the shot in visualizer.cofee. 

- New item in the settings to define a minimum shot duration in seconds to trigger the auto-upload. It defaults to 6 seconds because Ray has suggested a few times that best way to preheat the group is to start the espresso shot and stop it around 5 seconds.

- Doesn't upload shots done with profiles that have beverage_type values "cleaning" and "calibrate".

- upload and uploadShotData commands have been a bit remade so that meaningful errors are captured in the settings and can be shown to the user.

- New proc 'msg' is invoked whenever msg was being called, and just adds the plugin name to every message in the log file. This just a silly thing but I use the trick in all my plugins and find it useful for parsing the log.

- Added some [translate]s in borg toasts and plugin variables.

- web_browser in utils.tcl modified to open links on PC. Mac not supported as I can't test it. This was on John's TODO's on Diaspora upon my request, if approved here will inform him there that he doesn't need to do it.


Mimoja:
Manual rebase needed, doing that

Mimoja:
Merged

PR #85, opened 2021-03-11
DO NOT MERGE - Names of Insight graphs - candidate names

jeffsf:
DO NOT MERGE -- THIS IS FOR DISCUSSION OF THE "RIGHT" NAMES

Having names for the graphs will allow extensions to be able to
modify them, suvh as hiding elements, changing data sources,
or changing styling.


decentjohn:
This is a good point. I don't have any strong opinions on what the names should be.  My feeling is that anything that's consistent and obvious would be a massive improvement, and good enough.

So... use your judgement.

jeffsf:
Closed in favor of #86

PR #87, opened 2021-03-17
2021 03 17 resolve insight rename issue

jeffsf:


decentjohn:
confirmed, fixes all known issues.

PR #90, opened 2021-03-24
Convert bluetooth to MSGs with severity levels

Mimoja:

@jeffsf can I ask you to have a look?

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -41,9 +41,6 @@ proc scale_timer_start {} {
 }
 
 proc scale_timer_stop {} {
-
-	#msg "scale_timer_stop\n[stacktrace]"
-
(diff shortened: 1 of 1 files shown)


jeffsf (commented):
Great work!

I went cross-eyed trying to keep it all in perspective. I've got a few different opinions, but they're opinions.

I think that no matter what we do, we'll find reasons to adjust in the future. Getting these trimmed down enough so that normal users can run logging all the time at a moderate level (probably NOTICE and up) will, I think, help us all.

jeffsf:
The set of questions around what level to report attempts to communicate with an device that isn't connected still have me shifting back and forth. 

I finally came down to the effort in fixing the code wasn't worth it, nor was alarming users with a WARNING as a result of the code's higher level logic not checking that the device was connected before making a request.

PR #91, opened 2021-03-25
Add first pass at plugin documentation

sudowork:
Documented my findings based on reading the source.

### DIFF
--- a/documentation/de1_app_plugin_development_overview.md
+++ b/documentation/de1_app_plugin_development_overview.md
@@ -1,0 +1,131 @@
+# DE1 App Plugin (Extension) Development Overview
+
+This overview aims to provide an introduction to getting started with plugin (extension) development. It gives some background on how plugins and the plugin system works and gives some pointers to getting started with plugin development. While this document is a starting point, it does not aim to cover Tcl or how the rest of the DE1 application is architected. See [DE1 App Core: Shot Cycle Overview](de1_app_core_shot_cycle_overview.md) for more details on the operations of the DE1 tablet application.
+
+## Glossary
+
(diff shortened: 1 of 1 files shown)


Mimoja:
Extremely cool! I think it would be a good idea to always generate the settings from an array instead of using a default file, but i don't think those things are reflected in the example plugin yet and I can add them later to this doc

ebengoechea:
Nice @sudowork! This will be very helpful!

On the "Setup your development environment", I'd start by saying that downloading the DE1 app for your workstation (PC/Mac/Linux) from https://decentespresso.com/downloads is enough. it includes the right UndroidWish for your platform. You don't need to download it separatelly (but describing how to do it for customization is very useful).

AndroWish/UndroidWish [version](http://www.androwish.org/download/):
- Currently bundled with the app is the "Eppur si muove" version
- BEWARE latest version "The Flux Capacitor" does NOT work on android tablets if a bluetooth scale is used.
- Original DE1 tablets were released with version "The Leyden Jar". Many users have never updated (the DE1 app currently offers no automatic way of updating AndroWish), so the safe AndroWish version for development is "The Layden Jar".

Also important is to note that the "preload" command is **always** run, even if a plugin is not enabled. If an error takes place during the preload step, the app will be totally locked and the only way for users to get out of the locked app is to manually delete the plugin folder. So, the preload step should be **very** carefully analyzed and proof-tested by plugin authors.

ebengoechea:
I'll be documenting soon how the GUI layer of the app works, together with the new 'dui' package. Should be helpful for plugin writters too.

ebengoechea:
Also, you don't need to build **all** your GUI on the `preload` step, only the settings page, if you want it the "Settings" button to appear when your plugin is selected on the extensions page. The page name should be the return value of the preload command, and it's important to take into account that the settings page will be accesible even with your plugin disabled (for example, I disable several of the settings options in my plugins if the plugin is not enabled, as they only make sense when enabled). The rest of your GUI can (and should, in my opinion) be created in the `main` command.

decentjohn (approved):

ebengoechea:
By the way, we have just started documenting here after @jeffsf created the folder and added the first document a few days ago. I think it's great, and better to document technical stuff here than in Basecamp, but I personally would have used the Wiki feature of GitHub, instead of .md files on the repository. What are your feelings about what is best, wiki or .md's? @decentjohn @Mimoja @barneyhawes @sudowork

PR #93, opened 2021-03-25
Allow de1plus/unde1plus.sh to be launched from anywhere

sudowork:
Change working directory into the de1plus directory as the first part of the
script so that relative paths to *.tcl files work correctly.


Mimoja:
@decentjohn this is your territory, please have a look :)

decentjohn (approved):

PR #96, opened 2021-03-26
de1: Fix "negative time" display during some shots

jeffsf:
If a shot was interrupted during PreInfuse, the next shot would render
negative time in the GUI until the shot ended. Once that next shot ended,
the times rendered would be correct.

Traced to ::start_espresso_timers not properly resetting the timers.

Resolved by resetting those timers in that routine, presently in vars.tcl

There still is a behavior where, before flow begins, the times from
the previous shot appear. Changing this would mean that an "aborted"
shot would wipe the previous timers. Deferring any such changes to next
release cycle.

### DIFF
--- a/de1plus/vars.tcl
+++ b/de1plus/vars.tcl
@@ -331,6 +331,7 @@ proc stop_espresso_timers {} {
 }
 
 proc start_espresso_timers {} {
+	clear_espresso_timers
 	set ::timer_running 1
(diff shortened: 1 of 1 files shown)


Mimoja:
Looks good to me, thank you!

Mimoja:
Confirmed working with a water-only shot without pressure

PR #97, opened 2021-03-26
Resolve initialization-time race condition with scale, correct a comment

jeffsf:


jeffsf:
Corrected/clarified commit messages, no code changes.

PR #105, opened 2021-04-03
Remove "stale" profile and settings requests from BLE queue

jeffsf:
### This needs significant testing across more than just my machine before merge

Users have reported unresponsive BLE after interacting with profiles.
In one case, a queue depth of over 1300 was observed. Examination
of logs suggested that at least one cause was users updating profiles
and settings faster than they could be transferred to the DE1.

As the settings that may be pending in the BLE queue are about to be
overwrtten, remove "matching" entries prior to adding their replacements.

remove_matching_ble_queue_entries {comment_regexp} is used to match
against the "comment" that is provided to userdata_append

This is a "spot fix" and should be reexamined if the BLE queue
is rewritten in the future. The following calls are addressed:

  * de1_send_shot_frames
  * de1_send_steam_hotwater_setting
  * set_steam_flow
  * set_steam_highflow_start
  * set_tank_temperature_threshold

Additional calls may need to be addressed in a similar manner

NB: This manipulates the queue directly and is not thread-safe

    This is not deemed to be an issue at this time as the Tcl
    implementation is single-threaded.

    Should this change in the future, the queues will need mutex

### DIFF
--- a/de1plus/de1_comms.tcl
+++ b/de1plus/de1_comms.tcl
@@ -980,7 +980,15 @@ proc mmr_write { note address length value} {
 }
 
 proc set_tank_temperature_threshold {temp} {
+
+	###
+	### NB: The BLE queue is not thread safe
+	###
+
 	msg "Setting desired water tank temperature to '$temp' [stacktrace]"
+
+	catch { after cancel $::_pending_tank_temperature_change }
+	remove_matching_ble_queue_entries {^MMR set_tank_temperature_threshold}
 
 	if {$temp < 10} {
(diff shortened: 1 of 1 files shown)


Mimoja (approved):

Mimoja:
Tested on a v1.3 with MimojaCafe and kept queue length in check. yay! Log suggest proper acknowledgement of profiles if transmission is interrupted after the header. Doublechecking with @rheasman

Mimoja:
Should be fine from the DE1s firmwares perspective

PR #107, opened 2021-04-04
scale: Resolve logic issues with scale connecting during shot

jeffsf:
Users whose scales connected and began sending updates during a shot
were observing that SAW was not working as expected. Examination of
.shot files provided did not contain any raw scale data.

Issue was traced to the on_connect callback which was stopping recording
(which stops SAW) and resetting the shot record.

As these were already being handled at the start of Espresso or HotWater
(the two states that are recorded and to which SAW applies)
they were removed from the on_connect callback.

DEVELOPER NOTES
===============

  ::device::scale::history::on_state_change

renamed to

  ::device::scale::history::on_major_state_change

for consistency.

Version bump due to API change.


jeffsf:
Tested with Skale II, disconnected, connected, and multiple times powering it on after starting a shot.

All known issues resolved.

No Acaia or Felicita here to test with.

PR #111, opened 2021-04-06
Have Skins provide their own filelist to reduce the need to update mi…

Mimoja:
…sc.tcl

While at it: remove all the whitespaces that have gone rouge over the months

Every skin is scanned for a filelist.txt which will in return be used to create the new manifest. This means that
we decouple skin / submodule updates and App updates as there is no longer any need to touch misc.tcl

### DIFF
--- a/de1plus/misc.tcl
+++ b/de1plus/misc.tcl
@@ -98,12 +98,6 @@ proc make_de1_dir {srcdir destdirs} {
 
     }
 
-
-        # plugins/example.tcl *
-        # plugins/example-settings.tdb *
-
-
-
     set files {
         binary.tcl *
         bluetooth.tcl *
@@ -136,9 +130,9 @@ proc make_de1_dir {srcdir destdirs} {
         history_export.tcl *
         version.tcl *
         profile.tcl *
-	de1_de1.tcl *
-	device_scale.tcl *
-	event.tcl *
+        de1_de1.tcl *
+        device_scale.tcl *
+        event.tcl *
 
         profiles_v2/readme.txt *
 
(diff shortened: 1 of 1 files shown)


Mimoja:
@johnbuckman this needs your approaval
@jeffsf does this look sane to you?

jeffsf:
I'm going to put out a bias first, in that I wrote a script to scrape that list from the Tcl code to generate a list for rsync. You can't source `misc.tcl` and get the list.

I keep wondering why it isn't just a flat text list in each module and submodule (skin and extension both). If `misc.tcl` can generate what it needs from the flat files in the submodules, why not from the main module as well?

de1plus/misc.tcl
```diff
@@ -686,16 +662,26 @@ proc make_de1_dir {srcdir destdirs} {
...
         plugins/DGUI/setup_Insight.tcl *
         plugins/DGUI/setup_MimojaCafe.tcl *
+    }
 
+    # Have skins deal with their own filelist if they want to
+    set skin_folders [lsort -dictionary [glob -nocomplain -tails -type d -directory "[homedir]/skins" * ]]
+
+    foreach s $skin_folders {
+        set fbasename [file rootname [file tail $s]]
+        if {[file exists "[homedir]/skins/$fbasename/filelist.txt"] == 1} {
```
jeffsf, line 672: Personal preference as I start to get better with Tcl is to treat as a Boolean unless it is going to return something other than True False (or its equivalents). 

Here the meaning is clear. It is less clean with something like `$::settings(something) == 1` is that a Boolean, or what does "2" result in?

decentjohn:
I'm a bit nervous about simply including all files in git, in the distro.  There are likely files that are not meant for distro, just meant for dev.

How about there is a filelist.txt in the main de1plus dir, and in each extension and skin dir?

Another poss is to hav a proc that reads pkgIndex.tcl in each directory, and those are the files that get included in the filelist.txt.

de1plus/misc.tcl
ebengoechea: @jeffsf I tend to use [string is true $::setttings(something)], which takes 1, yes, true, or otherwise false.

PR #112, opened 2021-04-07
Revert "don't need to give an ENJOY message..."

jeffsf:
Contrary to the justification given for removal of the "Enjoy!"
message in commit caacd68, it indicates recording complete,
which is an important and distinct step in the sequence of a shot.

Further, the SAW or SAV messages ("Stopping for weight/volume") only
are presented when one of those two processes trigger. They are not
presented when a shot completes as it "walks off the end" of the steps.

This restores the intended, default behavior where the completion of
the shot is always indicated to the operator both visually and, for
users that rely on audio cues, wth "speak". By also indicting to the
user when SAV or SAW is triggered (visual only, as it is secondary
data) a user can now clearly distinguish between:

  * Shot terminated due to SAV
  * Shot terminated due to SAW
  * Shot terminated by the DE1 (likely end of profile)

This behavior resolved the long-standing complaints around the
messages in v1.34 and earlier being incorrect and improperly
indicating "weight reached" when SAW was not triggered.

Should an individual wish to remove or change the indication of this
important event on a shot's timeline, it can be overriden by a skin
or in an extension.

This reverts commit caacd68.

### DIFF
--- a/de1plus/gui.tcl
+++ b/de1plus/gui.tcl
@@ -3055,10 +3055,9 @@ namespace eval ::gui::notify {
 			}
 
 			record_complete {
-				# don't need to give an ENJOY message when the app already gives "stopped on volume, stopped on weight" messages just a few seconds earlier.
-				#set what [translate {Enjoy!}]
-				#borg toast $what
-				#say $what $::settings(sound_button_in)
+				set what [translate {Enjoy!}]
+				borg toast $what
+				say $what $::settings(sound_button_in)
 			}
 
 			saw_stop {


jeffsf:
Expanded commit message

decentjohn:
I am ok with a compromise with a "shot ended normally" message if the SAV/SAW was not triggered.

This final popup I find annoying for users of SAV/SAW.

Basically the problems for me are:
1) there are usually 2 messages for the end of each shot, which feel redundant
2) the final ENJOY is a meaningless message, it does not tell the user anything
3) the final ENJOY is displayed once the UI is ready to accept new taps, and is now in the way of fast use of the tablet.

Also, please, for civility, if you disagree with a change I make, please don't reverse it immediately.  First, post a message, and make your argument. Reversing my change feels quite aggressive, perhaps more so than you intended.

decentjohn:
If you agree with the above compromise, please change the ENJOY message to "Shot ended normally"

decentjohn:
@Mimoja pointed out to me that I did to you (reversing your code w/o discussion) exactly what I asked you to do.  She's right, my bad, apologies @jeffsf

jeffsf:
Alternate approach under test.

PR #115, opened 2021-04-10
Add local_time to .shot file

jeffsf:
The .shot file presently only contains a timestamp in seconds without
timezone.  Uploaded .shot files, such as from an extension, may no
longer have their name to assist in cross-referencing with the logs.

Add local_time as the line after "clock" at the top of the .shot file.


de1plus/vars.tcl
```diff
@@ -2957,6 +2957,7 @@ proc format_espresso_for_history {} {
 		set espresso_data {}
 		set espresso_data "name [list $name]\n"
```
Mimoja, line 2958: How does this work anyways!?

Mimoja (approved):

PR #117, opened 2021-04-12
logging: Disable logging when used during build

jeffsf:
_**Checked against app, needs to be checked against build environment there**_

As a result of inclusion by other packages, logging is initalized
during the build process. Console logging bleeds into the build logs.

To disable logging, set to Boolean True externally, prior to any
`package require` of de1app code. One way to do this would be:

    namespace eval ::logging { variable disable_logging_for_build True }

### DIFF
--- a/de1plus/logging.tcl
+++ b/de1plus/logging.tcl
@@ -55,6 +55,12 @@ namespace eval ::logging {
 
 	variable _log_fh ""
 
+	# To disable logging, set to Boolean True externally,
+	# prior to any `package require` of de1app code
+	#   namespace eval ::logging { variable disable_logging_for_build True }
+
+	variable disable_logging_for_build
+
 	proc safe_get {var_name} {
 
 		upvar $var_name var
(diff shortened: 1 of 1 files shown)


decentjohn:
firstly, thanks for teaching me this:
`[expr { [lindex [file split $argv0] end] in { "makede1.tcl" } }]`

I'd not seen that idiom before.

Now... let's say I'd like to be able to add my script to the list of no warnings, before sourcing this file. I don't see a way to do that. Something like this:

`[expr { [lindex [file split $argv0] end] in { [join "makede1.tcl" [ifexists ::logdisablelist]] } }]`

or does this work?

`set ::logging::disable_logging_for_build 1`

I think for that to work, you need to do an [info exists] test before redefining this variable, in case the calling script has already defined it.

jeffsf:
I think your idea of a variable you set in your "main" is a good one. 

I'm not a big fan of globals. Would you be OK with:
```
namespace eval ::logging { variable disable_logging_for_build True }
```
in your "main"?

---

I find the `in/ni` construct very useful especially when the list is expected to be changed, either with time, or at run time.

Don't like the choice of when your scale tares? No need to edit code, just change the list:
```
variable autotare_states [list "Espresso" "HotWater"]
```
```
	proc is_autotare_state {{state_text "None"}} {

		if { $state_text == "None" } { set state_text [::de1::state::current_state] }

		expr { $state_text in $::device::scale::autotare_states }
	}
```

decentjohn:
I like that:

`namespace eval ::logging { variable disable_logging_for_build True }`

decentjohn:
Confirmed to work 
https://github.com/decentespresso/de1plus_misc/blob/main/makede1.tcl

Issue #73, opened 2021-02-23
plugins: web_api: `package require wibble` causes errors, even if not enabled


jeffsf:
```
package require wibble
package require de1_machine

set plugin_name "web_api"
```
Causes failures on load in environments without `wibble` installed. If one is not going to use the plugin, there should be no reason to have to install `wibble`. Though preloaded in AndroWish, it is not present in, for example, macOS `wish`, including those installed with `brew`.

Suggest moving the offending dependency into a later phase of plugin initialization, where it is only evaluated if the plugin is active.

git merge-base main HEAD
55c65f15651aef5e8bb31180bb0bdd1f041bd162

jeffsf:
![Screen Shot 2021-02-23 at 10 51 41 AM](https://user-images.githubusercontent.com/1946364/108896825-3cdea580-75ca-11eb-8ea3-c250c3f42a20.png)

jeffsf:
Further, the attempts to define procs within the `::wibble` namespace on a global context also fail

Suggest that `namespace eval ::wibble` or similar be used.

![Screen Shot 2021-02-23 at 12 22 27 PM](https://user-images.githubusercontent.com/1946364/108903192-df4e5700-75d1-11eb-9e16-8697df683808.png)

jeffsf:
The loading problem seems to have been resolved by 495b831ac

The questions of where to declare ` proc ::wibble::check_auth` are primarily stylistic in this specific instance. They still are done outside of `proc main` and register both a namespace and procs.

PR #118, opened 2021-04-12
No preload

Mimoja:


jeffsf:
Would address #73 

_(I'll look at the code later on today)_

PR #114, opened 2021-04-09
Logging for v1.35

jeffsf:
The availability of logs during the lead-up to v1.35 has been very valuable to identify issues, be they with the code, the directions, the operator, or even the machine.

These commits assign priorities to hundreds of the previous `msg` command to make them "filterable" both for developers as well as for users and support-team members, properly provide 64 kB buffering to respect the concerns around flash wear, and enable logging for all users. Stable users will run "INFO" by default, with "DEBUG" available through a plug-in (or future skin modifications). Users on pre-Stable branches will log at the "DEBUG" level.

The "default" settings page has been modified to reflect that logging is on. This will help reduce confusion from experienced users expecting to see an option there, as well as less-experienced ones following outdated instructions.

_**These are "interim" severity assignments.**_

The eventual objective is that at the NOTICE level one gets a "quick sketch" of the operation of the machine. At the INFO level, there are enough details to determine, for example, when the scale tare was done, without overwhelming a non-technical users.

There's still a lot of "doubling" with both de1_comms and bluetorh reporting the same event. My plan in the future is to be able to report at the higher level (de1_comms) with a moderate log level and the lower level at a lower level. Users at the INFO level would then only see one of the pair of log messages.

Making the messages "denser" in content and eliminating non-printable "binary" seems needed to be able to achieve these objectives, as well as making it easier to select the "important enough" logging points for the NOTICE level and those that should be pushed down to the DEBUG level.

Thanks to @Mimoja and #90 for the contributions.

Many attempts to set bluetooth devices are done before the device is known to be connected. These, contrary to my instincts demanding checking _before_ they are called, are presently logged as DEBUG, though they arguably should be logged as a WARNING or ERROR in logic that let the call be made at all. Resolving that would require a rework of `bluetooth.tcl`.

No significant work around changing the _content_ of the messages is in this PR for v1.35. There has been significant work past this to reduce the size of the messages, eliminate "binary" data, and to make the messages more meaningful for a non-expert in that section of code. As these sometimes involve processing of the messages, especially within `bluetooth.tcl`, changes of this type are planned to be deferred to after the v1.35 release.

PR #119, opened 2021-04-14
Improve web_api plugin

jeffsf:
**This is not a priority for v1.35**

As noted in #73 the web_api plugin previously created and used the `::wibble` namespace, even if the plugin was not enabled. Resolved by moving the `proc ::wibble::*` definitions into `main {}`

Seeing duplication of the generation the response object and expecting to add to the API, made generalized "200"  returns for JSON and text data.

In resolving that, noted that the port was hard-coded.

Added a /flush action, which calls `::logging::flush_log` (Introduced by #114, not merged at this time).

### DIFF
--- a/de1plus/plugins/web_api/plugin.tcl
+++ b/de1plus/plugins/web_api/plugin.tcl
@@ -126,7 +126,7 @@ namespace eval ::plugins::${plugin_name} {
         # Start a server and enter the event loop if not already there.
 
         catch {
-            ::wibble::listen 8080
+		::wibble::listen $::plugins::web_api::settings(webserver_port)
         }
 
 	}  ;# main
--- a/de1plus/plugins/web_api/plugin.tcl
+++ b/de1plus/plugins/web_api/plugin.tcl
@@ -99,6 +99,16 @@ namespace eval ::plugins::${plugin_name} {
 		::wibble::return_200_json
 	}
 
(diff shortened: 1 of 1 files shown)


de1plus/plugins/web_api/plugin.tcl
```diff
@@ -34,84 +33,112 @@ namespace eval ::plugins::${plugin_name} {
...
+	}
 
-proc ::wibble::togglePowerOn {state} {
-    if { ![check_auth $state] } {
-        return;
-    }
+	# Utilities
 
-    start_idle
+	proc ::wibble::return_200_json {{content "{status: ok}"}} {
```
Mimoja, line 64: Something seems a bit off with your whitespaces here

Mimoja (approved):

de1plus/plugins/web_api/plugin.tcl
jeffsf: New code: Single tab indentation, consistent with much of the de1app Tcl code I've worked with.
`start_idle`: 4 spaces

We probably should come up with a standard (my preference including no trailing whitespace and each file ends with a newline with no blank lines)

I can see arguments for or against spaces/tabs and 4/8-column indentation

jeffsf:
Corrected MIME type on `return_200_text` to `{text/plain charset utf-8}`

PR #123, opened 2021-04-17
scale: Improve default wording on timeout toast

jeffsf:
_**Note: Content changed to provide a "back to normal" message. See commits of 2021-04-19**_

_**Non-functional change. Not essential for v1.35. See DD To-dos.**_

If the scale code does not see updates from the scale for longer than
one second, it informs the user. In most cases during "normal"
operation, this is due to a long-running process in another part of
the app taking to long to relinquish execution.

At present, there is no facility for informing the user that normal
updates have resumed. There are toasts raised at the beginning of a
shot if it appears that the user intends for SAW to be operational and
the scale is either disconnected or not updating.

As this timeout is generally transient in nature, clarify the wording
to indicate the action the user needs to take, "Check scale".


ebengoechea:
@jeffsf Wouldn't it make sense to also make the interval longer than 1 second, to prevent "false" messages in those cases you mention of long-running processes?

jeffsf:
@ebengoechea 

With BLE able to run at 10 ms per transaction, running around 80 ms median on my v1.3 tablet and the current app, 1 second is already "forever". 

Anything that is taking over a few hundred ms really should be rewritten to be split up to allow other, time-critical processes to run. There's no preemption in a Tcl interpreter. Once the BLE queue backs up, something like a "Stop" command can take two seconds to be processed.

ebengoechea:
@jeffsf If something "has to be rewritten", documenting why and how is necessary. I don't even know how to check if my own code needs that or not, or how would I need to change it. For example the SDB plugin may likely block during its sync step at startup, but right now I have no idea how to check or how to change it.

Mimoja:
Maybe "Please Check Scale" would sound just that tiny bit better?

decentjohn:
Simplest msg is likely "Scale disconnected" and if a "Scale connected" toast also pops up, this isn't bad at all.

jeffsf:
I'll see what I can do tomorrow for a toast for "Scale running" or "Scale updating" 

This isn't completely right with logic, but a quick hack to see what you think about it.
```
--- a/de1plus/device_scale.tcl
+++ b/de1plus/device_scale.tcl
@@ -386,7 +386,8 @@ namespace eval ::device::scale {
                }
 
                if { $::device::scale::_watchdog_entry_id  == "" } {
-                       msg -DEBUG "Scale watchdog first starting with handle $::de1(scale_device_handle)"
+                   msg -DEBUG "Scale watchdog first starting with handle $::de1(scale_device_handle)"
+                   borg toast "Scale updating"
                } else {
                        after cancel $::device::scale::_watchdog_entry_id
                }
```

The scale usually isn't disconnected. `ble` still thinks it's there, and it likely is. Most of the time it is the code taking far too long to execute. Get rid of the long-running code and this generally becomes a moot point -- if it occurs, your scale is likely having problems, not the rest of the app.

Anything that takes more than a couple hundred ms should really be split up. I'd typically say 50 ms or less if you want 100 ms latency, but things are generally slow on the tablet. 

Without knowing where the slow spots in the code are, it's hard to say. One common task that can be split up is finding a list of things and then processing them. Let's say it takes 20 ms to create the list, and 2 ms per item. 10 items in 40 ms, 100 items in 220 ms, 1000 items in 2 seconds, ...

Rather than something like
```
foreach item [generate_list] {
    process item
}
```
assuming that the Tcl event queue works well, conceptually you can do, if time critical
```
set the_list_to_process [generate_list]
if { [llength $the_list_to_process] > 0 } {
    after 1 [list process_next $the_list_to_process]
}

proc process_next (item_list} {
    this_item = [lindex item_list 0]
    remaining_items = [lrange item_list 1 end]
    process this_item
    if { [llength $remaining_items] > 0 } {
        after 1 [list process_next $remaining_items]
    }
}
```
I'm assuming that `after 1` won't move ahead of other items on the event queue that haven't been processed. I'm less trusting of `after 0`

If not time critical, you can add a little delay, or, if possible, use `after idle`

If you thrown them at the queue all at once, nobody else gets in with their events.

You should be able to use the same concept to split up
```
fast_call
this_takes_a_while
and_this_does as well
another_fast_call
```
by giving a moment's break at least between the two long-running calls, if not before and after as well.

decentjohn:
I don't have a problem with spinning the BLE handler into its own thread, for the next beta cycle:
https://www.tcl.tk/man/tcl/ThreadCmd/thread.htm

With extensions coming in, I think that timing is going to continue to be a difficult problem for BLE at 10hz.

ebengoechea:
@jeffsf Thanks for the details. SDB/DYE plugins do such long ops when synchronizing the shot history database with the database, normally only during startup, but can also be triggered manually from its settings pages. I’ll work on changing them as you say.

jeffsf:
Converted to draft pending adding a "scale reporting" indication once the scale updates start processing again

jeffsf:
Checked for happy path, scale disconnect -- both work as expected

With long-running tasks, sometimes they are flagged, sometimes not. After beating on this without a debugger for a couple hours, from what I can tell, it is that there are sometimes pending updates on the event queue that reset the single watchdog now in use. This is consistent with previous behavior.

Below, caught just as the message appears after DSx takes execution for several seconds, with the "Scale reporting" message overriding the "Check scale" message.

I struggled for the right wording before settling on "reporting". 
- Scale updating -- I didn't ask to flash firmware, did I?
- Weight updating -- Hmm, still says "0.0". Should I wait for it to finish?
- Scale running -- sort of goofy
- Scale OK -- It had better be more than OK, I paid over $200 for it
- Scale connected -- It never disconnected


Screenshot_20210419-193726

decentjohn:
My own feeling is that "scale connected" is clear enough, but if you want to differentiate connected vs weighing, I am ok with

Scale connected
Scale disconnected
Scale is working
Scale is not working

jeffsf:
It's the skin or extension that is broken, not the scale, in most cases. 

The scale has been _connected_  all along. You don't need to go to the setup page and reconnect it. 

The scale has _working_ all along. it's just that somebody's skin or extension is hogging the tablet.

decentjohn:
ok

PR #126, opened 2021-04-24
de1: Protect against unrecognized states and substates

jeffsf:
User reported a condition where, when using a refill kit, after
starting the DE1 but before the tray had been filled, a substate of
decimal 20 was reported. As this is not in the arrays in machine.tcl,
an error occurred. As the current state and substate are often checked
on each screen refresh, the GUI became non-responsive.

Provide a mitigating fix by capturing and logging the lookup failure.
As the substate is not known to the app or skin code, this should not
impact any exisitng functionality.

suggest backport to stable.

Resolve further in the future with updates to machine.tcl for any
known-missing states and substates.

### DIFF
--- a/de1plus/de1_de1.tcl
+++ b/de1plus/de1_de1.tcl
@@ -421,12 +421,20 @@ namespace eval ::de1::state {
 
 	proc current_state {} {
 
-		expr { $::de1_num_state($::de1(state)) }
+	    try { set state_text [expr { $::de1_num_state($::de1(state)) }] } on error result {
+		set state_text [format "unknown_0x%02x" $::de1(state)]
+		msg -ERROR [format "Unrecognized API_MachineStates 0x%02x" $::de1(state)]
+	    }
+	    return $state_text
 	}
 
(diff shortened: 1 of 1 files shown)


jeffsf:
I have tested "happy path" but can't replicate the reporter's situation.

Tested in isolation for "error" condition:
```
(BLE_Code) 124 % array set de1 {state 10 substate 4}
(BLE_Code) 125 % current_state
Descale
(BLE_Code) 126 % current_substate
preinfusion
(BLE_Code) 127 % array set de1 {state 30 substate 20}
(BLE_Code) 128 % current_state
-ERROR Unrecognized API_MachineStates 0x1e
unknown_0x1e
(BLE_Code) 129 % current_substate
-ERROR Unrecognized API_Substates 0x14
unknown_0x14
```

jeffsf:
Switching back to draft with the new information that there is no substate 20!

jeffsf:
Looks like the root cause is not missing decoder-ring entries. Withdrawing at this time and retaining the branch locally.

PR #127, opened 2021-04-24
Utilize the build-info file in the app

jeffsf:
Ref: https://github.com/decentespresso/de1plus_misc/pull/2

_**This should get a version version bump for `de1app.tcl` and be reflected in `skins/default/de1_skin_settings.tcl`**_

Read in the build-info file if present, log it at the top of the log (so logs uploaded always have detailed app-version information).

Provide programmatic access to the build-info contents.

~~Use the more-detailed string in the GUI~~

**No GUI changes**

### DIFF
--- a/de1plus/misc.tcl
+++ b/de1plus/misc.tcl
@@ -141,6 +141,8 @@ proc make_de1_dir {srcdir destdirs} {
         device_scale.tcl *
         event.tcl *
 
+	build-info *
+
         profiles_v2/readme.txt *
 
         history/info.txt *
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,8 @@ _$*
 core
 # CVS default ignores end
 
+/de1plus/build-info
+
 release
 de1plus/plugins/visualizer_upload/settings.tdb
 de1plus/plugins/web_api/settings.tdb
--- a/de1plus/de1app.tcl
+++ b/de1plus/de1app.tcl
@@ -18,6 +18,93 @@ try {
 	exit
 }
 
+#
+# Inline for now, then move out
+#
+
+namespace eval ::app {
+
+	variable build_info_filename "build-info"
+	variable build_info
+	set build_info [dict create]
+
+	variable build_string ""
+	variable build_timestamp 0
+
+	proc load_build_info {} {
+
+		variable build_info_filename
+		variable build_info
+
+		if { [file readable $build_info_filename] } {
+
+			set _fh [open $build_info_filename "r"]
+
+			foreach _line [split [read $_fh] "\n"] {
+
+				msg -NOTICE "build-info: $_line"
+
+				if { [string length $_line] == 0 } { continue }
+				if { [regexp {^[:space:]*#} $_line] } { continue }
+
+				set _kv [split $_line "\t"]
+				set _k [lindex $_kv 0]
+				if { [llength $_kv] == 1 } {
+					set _v ""
+				} else {
+					set _v [join [lrange $_kv 1 end] "\t"]
+				}
+
+				dict append build_info $_k $_v
+			}
+		} else {
+
+			msg -WARNING "build-info: No such file:" \
+				$build_info_filename
+		}
+	}
+
+	proc ensure_build_strings {} {
+
+		if { [dict exists $::app::build_info version_string] } {
+			set ::app::version_string [dict get $::app::build_info version_string]
+		} else {
+			set ::app::version_string [package version de1app]
+		}
+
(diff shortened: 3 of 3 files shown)


jeffsf:
## Stable

Screenshot_20210424-134050

## Beta

Screenshot_20210424-140452


## Nightly

### Before any tagged Betas:

Screenshot_20210424-134532

### After tagged Beta:

Screenshot_20210424-134912

jeffsf:
Replaced my "local" choice of `BUILD INFO:` with the more grep-friendly `build-info` after looking at the previous log output.

de1plus/misc.tcl
```diff
@@ -140,6 +140,9 @@ proc make_de1_dir {srcdir destdirs} {
         de1_de1.tcl *
         device_scale.tcl *
         event.tcl *
+	event.tcl *
```
Mimoja, line 143: There seems to somewthing wrong here

de1plus/misc.tcl
```diff
@@ -140,6 +140,9 @@ proc make_de1_dir {srcdir destdirs} {
...
+
+	build-info *
```
Mimoja, line 145: Please create a placeholder file so that a clean clone is always a full app

Mimoja (changes requested):

de1plus/misc.tcl
jeffsf: Confirmed that logic checks for the file being readable before trying to open it. As a result, this will function if `build-info` is not present on the user's machine.

rheasman:
I can't emphasize enough how desperately the DE support people need this. Adding the app version in the information box that describes the hardware and firmware versions would be a lifesaver, so we can ask for a single screenshot. Right now, I am dealing with "Well, it was the beta version, about 2 weeks ago, maybe?" as a version. And "stable, um, a little while ago, I think?" as app versions.

Damian-AU:
DSx does

jeffsf:
For any changes to Insight past that simple variable swap, you'll have to work with John directly.

ebengoechea:
@rheasman Something like this? Easy to add if @decentjohn agrees.

image

decentjohn:
I'm perfectly happy to have more stuff in the VERSION field, and key=value fields, comma separated, is the way to do it.  That ui field has two lines and is set to wrap.  Just modify the Tcl proc that creates this text string.

decentjohn:
Regarding what I did in combing "track" and "version" into a single thing:

Screen Shot 2021-05-06 at 11 26 40 AM

I'm not sure it was a good idea now, because 
a) when you change track
b) the label here changes immediately
c) and also after a release, the beta/nightly/stable tracks all have the same version, yet are labelled differently

and so I'm not sure it's a good idea to build on top of that mistake, to build the build number too:
Screen Shot 2021-05-06 at 11 26 05 AM

Perhaps this is a good time to propose a better way to indicate /track/version/build in the UI?

I've seen builds in parens, which is not a bad thing, so perhaps this would be ok?

`v1.36.8-13 (GF6844170) "nightly"`

de1plus/misc.tcl
Mimoja: the updater will not create the distro and the download will fail if not present on the server

ebengoechea:
@rheasman @decentjohn @jeffsf 
image

jeffsf:
I don't really care how the characters are shown on the screen as long as it reveals a uniquely identifiable set of sources.

Have at it. Tcl has regex. 

The problem is that reports of issues to either the official of volunteer support teams are meaningless, even with "I'm on stable". 

There have already been issues where different builds have been released with the same version number. Part of this is process improvement, which is why the scripts create the `version.tcl` and `build-info` directly from _annotated_ tags. I can't wire that in.

As John points out, stable, beta, nightly are simply a setting for where to obtain updates. It isn't a reliable indicator of what is on the machine. You can even get the same build from multiple distribution points.

jeffsf:
_**No changes to the GU I in this PR.**_

Since the presentation layer should be kept separate from the operational layer, someone else can deal with it.

de1plus/misc.tcl
jeffsf: Sounds like a build system problem to me. `build-info` gets created, along with the proper `version.tcl` file. Both are in the list of files to be included. 

What kind of build system can't handle dynamically generated artifacts?

decentjohn:
I'm happy to have this PR approved @Mimoja if there's no UI aspect to it, and I can do this:

`v1.36.8-13 (GF6844170) "nightly"`

as a separate commit, and we can change our minds about whether we like that or not, but at least it'll be an improvement.

Issue #129, opened 2021-04-27
Constant message about scale being disconnected


oscar-b:
"Warning: Scale not connected" is displayed every 30s, even when the machine is in sleep and displaying the screen saver. This message also causes the tablet to make a ticking sound.

```
2021-04-27 09:09:40.368 NOTICE: acaiascale disconnected handle ble81 address 00:1C:97:17:FE:DE rssi -150 state disconnected
2021-04-27 09:09:40.393 NOTICE: Connecting to scale on 00:1C:97:17:FE:DE
2021-04-27 09:09:40.400 DEBUG: Scale watchdog cancel - no ID to cancel
2021-04-27 09:10:10.407 NOTICE: acaiascale disconnected handle ble82 address 00:1C:97:17:FE:DE rssi -150 state disconnected
2021-04-27 09:10:10.436 NOTICE: Connecting to scale on 00:1C:97:17:FE:DE
2021-04-27 09:10:10.439 DEBUG: Scale watchdog cancel - no ID to cancel
2021-04-27 09:10:40.449 NOTICE: acaiascale disconnected handle ble83 address 00:1C:97:17:FE:DE rssi -150 state disconnected
2021-04-27 09:10:40.471 NOTICE: Connecting to scale on 00:1C:97:17:FE:DE
2021-04-27 09:10:40.473 DEBUG: Scale watchdog cancel - no ID to cancel
2021-04-27 09:11:10.479 NOTICE: acaiascale disconnected handle ble84 address 00:1C:97:17:FE:DE rssi -150 state disconnected
2021-04-27 09:11:10.506 NOTICE: Connecting to scale on 00:1C:97:17:FE:DE
2021-04-27 09:11:10.508 DEBUG: Scale watchdog cancel - no ID to cancel
2021-04-27 09:11:40.515 NOTICE: acaiascale disconnected handle ble85 address 00:1C:97:17:FE:DE rssi -150 state disconnected
2021-04-27 09:11:40.543 NOTICE: Connecting to scale on 00:1C:97:17:FE:DE
2021-04-27 09:11:40.544 DEBUG: Scale watchdog cancel - no ID to cancel
2021-04-27 09:12:10.554 NOTICE: acaiascale disconnected handle ble86 address 00:1C:97:17:FE:DE rssi -150 state disconnected
2021-04-27 09:12:10.582 NOTICE: Connecting to scale on 00:1C:97:17:FE:DE
[...]

PR #131, opened 2021-04-27
gui: Suppress scale disconnect messages when DE1 is in Sleep

jeffsf:
User reported that every 30 seconds, evne when sleeping, the toast
"Warning: Scale not connected" is displayed, along with a ticking sound.

This can be replicated by configuring a scale, turning it off, and setting
automatically_ble_reconnect_forever_to_scale 1

It is due to the `ble` stack sending a connection event as it starts
to try to connenct by address and then a disconnect event when it fails.

Gate the message so it is not displayed when the DE1 is in Sleep
state.

Long-term resolution is a BLE stack that properly reports
connection state.

Resolves https://github.com/decentespresso/de1app/issues/129

### DIFF
--- a/de1plus/gui.tcl
+++ b/de1plus/gui.tcl
@@ -3013,9 +3013,16 @@ namespace eval ::gui::notify {
 
 			not_connected {
 
-				set what [translate {WARNING: Scale not connected}]
-				borg toast $what
-				say $what $::settings(sound_button_in)
+			    # With automatically_ble_reconnect_forever_to_scale 1
+			    # `ble` will report a connection event when attempting to connect.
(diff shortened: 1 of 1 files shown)


Mimoja (approved):

jeffsf:
Not still in progress -- more that _when_ I get something to better track Bluetooth state that it might be redundant.

I've been thinking about that and I can see a scale coming and going during the night being annoying, so will probably leave this to handle that case in the future

oscar-b:
@jeffsf wouldn't it make sense to disable reconnections in the first place, even with `automatically_ble_reconnect_forever_to_scale` active, when in sleep mode?

PR #132, opened 2021-04-28
profiles: File name reflects profile name

lessthanjoey:
Users have difficulty managing profile files as they are named with timestamps that don’t give a clue to their contents.

Modified save_profile to use a name derived from the profile_title. Name
made “safe” by replacing whitespace with underscores, changing / to a
doubled underscore, and removing other potentially unsafe symbols and
unicode control characters

Signed-Off-By: Joe Duncan <j@josephd.net>

### DIFF
--- a/de1plus/vars.tcl
+++ b/de1plus/vars.tcl
@@ -2831,16 +2831,23 @@ proc save_profile {} {
 	} else {
 		# if they change the description of the profile, then save it to a new name
 		# replace prior usage of unformatted seconds with sanitized profile name and append with formatted time if file exists
+		
 		set profile_filename $::settings(profile_title)
+		set profile_timestamp [clock format [clock seconds] -format %Y%m%d_%H%M%S] 
 		regsub -all {\s+} $profile_filename _ profile_filename 
 		regsub -all {\/+} $profile_filename __ profile_filename 
 		regsub -all {[\u0000-\u001f;:?<>(){}\[\]\|!@#$%^&*-\+=~`,.'"]+} $profile_filename "" profile_filename
+		set profile_filename [string range $profile_filename 0 59]
 		if {[file exists "[homedir]/profiles/${profile_filename}.tcl"] == 1} {
-			append profile_filename "_" [clock format [clock seconds] -format %Y%m%d_%H%M%S] 
+			append profile_filename "_" $profile_timestamp
 			}
 	}
 	
 	set fn "[homedir]/profiles/${profile_filename}.tcl"
+	
+	if {[write_file $fn ""] == 0} {
+		set fn "[homedir]/profiles/${profile_timestamp}.tcl"
+		}
 
 	# set the title back to its title, after we display SAVED for a second
 	# moves the cursor to the end of the seletion after showing the "saved" message.
--- a/de1plus/vars.tcl
+++ b/de1plus/vars.tcl
@@ -2830,7 +2830,14 @@ proc save_profile {} {
 		set profile_filename $::settings(profile_filename) 
 	} else {
 		# if they change the description of the profile, then save it to a new name
-		set profile_filename [clock seconds]
+		# replace prior usage of unformatted seconds with sanitized profile name and append with formatted time if file exists
+		set profile_filename $::settings(profile_title)
+		regsub -all {\s+} $profile_filename _ profile_filename 
+		regsub -all {\/+} $profile_filename __ profile_filename 
+		regsub -all {[\u0000-\u001f;:?<>(){}\[\]\|!@#$%^&*-\+=~`,.'"]+} $profile_filename "" profile_filename
+		if {[file exists "[homedir]/profiles/${profile_filename}.tcl"] == 1} {
+			append profile_filename "_" [clock format [clock seconds] -format %Y%m%d_%H%M%S] 
+			}
 	}
 	
 	set fn "[homedir]/profiles/${profile_filename}.tcl"


Damian-AU:
Please do not rename my profiles, it will break DSx functions

lessthanjoey:
This is just changing the filenames of _newly_ saved profiles to not use the seconds timestamp. Anything that is currently saved with a name will keep that name.

decentjohn:
so... I previously did this, and had big problems with non-English languages and just, in general, making filenames that worked everywhere.  Different OSes have different unliked characters.

Can you stick to a passthrough filter that ONLY lets through a-z  and all other characters are blocked (space becomes _)

And can I ask for you to add a test first that the filename can actually be created and [file exists] afterwards, and if that fails, you fallback to a timestamp?

I had previously tried to make this work and failed.  It caused lots of bugs.

Mimoja:
@lessthanjoey Given Johns comment, do you plan on updating this PR once way or another or just giving it a tought if you think he is wrong?

lessthanjoey:
I've been giving it some thought :).

@decentjohn  I think it's a good idea (and easy enough) to add the file creation test with fallback. Regarding the whitelisting vs blacklisting, I actually initially implemented what you're asking but then @jeffsf pointed out that it'd effectively exclude all international customers from being able to take advantage of it and I wanted to work around that - even a rename of the built-in Rao Allongé profile to something like "Mine/Rao Allongé" would drop the é unless we also whitelisted that and it kinda spirals out quickly. So, *do you think the file creation test with fallback to a timestamp effectively handles your concern about the blacklist being accurate/sufficient for now?* Or do you want to go forward with the whitelist that's primarily focused on English? I think the feature has value even in the latter case but also thought there was value in trying to be more inclusive of global customers.

jeffsf:
I'm curious as to which OSes utf-8 filenames are a problem with.

If I were a non-American, I'd be somewhere from partially offended (Latin character sets) to completely offended that _my_ language was brutally stripped out.

At least as far as I knew, any recent version of Android, macOS, Linux, or Windows supported them (with the usual restrictions on path separators and the unwise use of "shell specials").

Older versions of Windows 10 (prior to 1607) still have a 260-character path limit, so it would probably be wise to limit the name to a reasonable length. I've got one that has a 55-character `profile_title` -- `jmk/EB88 Easy blooming 88°C - pressure dynamic decline`

One way to do it is to pick a number and leave enough for `_1620425078.tcl` -- if you say the total can't exceed 80 characters, that's 65 for the truncated name.

macOS:
```
[jeff@mini /tmp]$ touch 完成した水
[jeff@mini /tmp]$ ls -l 完成した水
-rw-r--r--  1 jeff  wheel  0 May  7 15:06 完成した水
```

Debian
```
jeff@pi-walnut:/tmp $ touch 完成した水
jeff@pi-walnut:/tmp $ ls -l 完成した水
-rw-r--r-- 1 jeff jeff 0 May  7 15:07 完成した水
```

Android on v1.3 tablet
```
em_t8880_64mme_jinsen:/sdcard # touch 完成した水
em_t8880_64mme_jinsen:/sdcard # ls -l 完成した水
-rw-rw---- 1 root sdcard_rw 0 2021-05-07 15:08 完成した水
```

lessthanjoey:
New commit adds file creation test with fallback to (formatted) timestamp if the file creation fails, as well as truncating the filename to fit within 80 chars.
File creation fallback tested by temporarily commenting-out the timestamp uniquifying code, setting default.tcl to read-only, and trying to save a profile called "default~". This was correctly saved as <timestamp>.tcl instead. Normally this would be solved as a uniquified default_<timestamp>.tcl.

I'll wait on @decentjohn 's further thoughts before flipping the blacklist to a whitelist.

decentjohn:
Thanks for doing that!

I'll pass this onto @Mimoja for approval now, as feature wise it now does what I'm comfortable with.

decentjohn:
As far as going beyond ASCII, my experience is that it causes so many problems, I prefer not to.  Anyone who feels differently can rename their files themselves, inside a file manager.

decentjohn:
The problem you are facing here is one I have already faced on http://magnatune.com where I make mp3 filenames from the song metadata.  Over the years, I found that ASCII was the only set of characters that didn't create annoying problems somewhere.  I don't really think this feature you've done needs to be more complicated than taking a stab at a recognizable filename.



    regsub -nocase -all "&ccedil;" $bio "c" bio
    regsub -nocase -all "&aelig;" $bio "ae" bio
    regsub -nocase -all "&ntilde;" $bio "n" bio
    regsub -nocase -all "&amp;" $bio " and " bio
    regsub -nocase -all "&atilde;" $bio "a" bio
    regsub -nocase -all "&igrave;" $bio "i" bio
    regsub -nocase -all "&egrave;" $bio "e" bio
    regsub -nocase -all "&sup1;" $bio "1" bio
    regsub -nocase -all "&sup2;" $bio "2" bio
    regsub -nocase -all "&eacute;" $bio "e" bio
    regsub -nocase -all "&euml;" $bio "e" bio
    regsub -nocase -all "&ecirc;" $bio "e" bio
    regsub -nocase -all "&acirc;" $bio "a" bio
    regsub -nocase -all "&oslash;" $bio "o" bio
    regsub -nocase -all "&amp;" $bio "&" bio
    regsub -nocase -all "&iuml;" $bio "i" bio
    regsub -nocase -all "&icirc;" $bio "i" bio
    regsub -nocase -all "&iacute;" $bio "i" bio
	regsub -nocase -all "&uacute;" $bio "u" bio
	regsub -nocase -all "&aacute;" $bio "a" bio
	regsub -nocase -all "&agrave;" $bio "a" bio
    regsub -nocase -all "&aring;" $bio "a" bio
    regsub -nocase -all "&bull;" $bio "*" bio
    regsub -nocase -all "&auml;" $bio "a" bio
    regsub -nocase -all "&ouml;" $bio "o" bio
    regsub -nocase -all "&ocirc;" $bio "o" bio
    regsub -nocase -all "&uuml;" $bio "u" bio
    regsub -nocase -all "&oacute;" $bio "o" bio
    regsub -nocase -all "&ograve;" $bio "o" bio
    regsub -nocase -all "&nbsp;" $bio " " bio
    regsub -nocase -all "&szlig;" $bio "ss" bio
    regsub -nocase -all "&#382;" $bio "z" bio
    regsub -nocase -all "&#352;" $bio "S" bio
    regsub -nocase -all "&#353;" $bio "s" bio
    regsub -nocase -all "&#269;" $bio "c " bio
    regsub -nocase -all "&#263;" $bio "c " bio
    regsub -nocase -all "&#402;" $bio "f" bio
    regsub -nocase -all "&#345;" $bio "r" bio
    regsub -nocase -all "&#328;" $bio "n" bio
    regsub -nocase -all "&#333;" $bio "o" bio
    regsub -nocase -all "&reg;" $bio "(r)" bio

PR #141, opened 2021-04-30
Fixes errors in dui sound. Warns if sound undef or not found.

ebengoechea:
This corrects some errors in the dui sound subensemble that made sounds not sounding. Also, now warns if a sound is not defined or the sound file path is not found.


ebengoechea:
Main change in this PR was included in #136 which has been already merged

PR #144, opened 2021-05-01
New dui subensemble 'dui image'

ebengoechea:
Image commands in DUI are now grouped under the subensemble namespace 'dui image':
- New commands 'dui image dirs', 'dui image add_dirs', 'dui image add', 'dui image find' and 'dui image photoscale'
- Resize images dynamically if needed on 'dui add image'
- De-hardcode the base screen size 2560x1800 everywhere.
- When resizing PNG files, now transparent background is preserved.
- Don't fail if an image file is not found, just warn in the log.
 
In addition, I made the following changes:

- Add DUI & platform data to the log on init to facilitate debugging.
- Fix bug introduced by first version of DUI, which was storing in the settings the variable "screen_size_height)"
- Make sound_name in 'dui say' optional.
-


Mimoja (approved):
Testing in nightly outstanding

PR #145, opened 2021-05-02
Theme new history viewer. Get DUI to style graphs

ebengoechea:
This styles all visual items in the new history viewer using DUI, so its aspect now adapts to the skin in use (if the skin provides its own DUI theme). Defining a general DUI skin theme should be enough to get the History Viewer to match the visual aspect of the skin, but further customization can be done modifying the desired aspects of the "hv_*" styles for the skin theme (hv_button, hv_done_button, hv_graph, etc.)

Several changes were done in DUI itself to allow this styling:

- dui::add:listbox gets a new -select_cmd option to set the callback to run when an item is selected in the listbox. No need to bind it manually now.
- dui::aspect::list gets a new -as_option option that allows to return a list of aspects to be used directly in a widget creation or configuration call. Useful to style graph lines and axis.
- now aspects whose value is an empty string can be defined. Before they were interpreted as missing and search continued. Allowing them permits to "reset" an option in a "substyle".
- new options updated in DUI documentation

### DIFF
--- a/de1plus/dui.tcl
+++ b/de1plus/dui.tcl
@@ -678,7 +678,7 @@ namespace eval ::dui {
 			default.dbutton.fill "#c0c5e3"
 			default.dbutton.disabledfill "#ddd"
 			default.dbutton.outline white
-			default.dbutton.disabledoutline "pink"
+			default.dbutton.disabledoutline "#ddd"
 			default.dbutton.activeoutline "orange"
 			default.dbutton.width 0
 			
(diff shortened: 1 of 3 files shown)


ebengoechea:
PR updated with a new commit that:
- Replaces $::iconik_settings(max_history_items) by a hardcoded "100" (at the moment). Needed to run with skins different than MimojaCafe.
- Allows bidirectional parameter passing between the history_viewer page and the invoking page (without using global variables). The load and init commands now take opening parameters (only implemented at the moment is -callback_cmd) and when "Done" is clicked control is returned to the invoking page (instead of the hardcoded previous "off") or the callback command is evaluated if it was an open parameter. In this case, 2 arguments are appended to the command, giving the clocks of the shots selected on the left and right listboxes. This data is already used by the DYE plugin.

Mimoja (approved):

PR #148, opened 2021-05-04
de1: scale: Track Bluetooth connectivity state better

jeffsf:
Only consider the DE1 or the scale to be "connected"
if discovery has been completed.

Note that this ia a change from all prior code where the presence
of a handle from `ble` indicated "connected". This handle would be
present during connect-by-address attempts, even if they failed to
find the device.

Only fire on_disconnect events if the DE1/scale was connected,
using the newly implemented definition.

Adds ::de1::is_connected predicate

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -847,13 +847,13 @@ proc check_if_initial_connect_didnt_happen_quickly {} {
 	set ble_scan_started 0
 	if {$::de1(device_handle) == 0 } {
 		catch {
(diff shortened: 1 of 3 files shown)


jeffsf:
Putting this on hold until I can confirm it won't change with current work with rewriting the BLE queues.

jeffsf:
Withdrawing this as it would be changed in potential, upcoming rewrite of the BLE connection, discovery, and queueing mechanisms.

It provides little, if any, benefit with the current code.

PR #155, opened 2021-05-14
Metadata package 

ebengoechea:
An initial simple implementation of a metadata package as described [on Basecamp](https://3.basecamp.com/3671212/buckets/7351439/messages/3762617806):

Having a data dictionary for the application fields in place should help in several ways:
- **Avoid code repetition**. Many values, such as lower and upper limits for variables, are currently hardcoded in many places in the code, which leads to problems (how often we've seen "probably that bug is due to a GUI control somewhere setting a limit on the variable, which is constraining it everywhere"?). Having a standard way to retrieve those values should help mitigate the problem. And anything that avoids code repetitition ("copy & paste") helps code maintenance.
- **Provide a standard way to add new fields**. Adding new data has proved to be a source of bugs, as often they are not reset properly. This is a problem specially when the data is introduced by skins and plugins/extensions, as even if the skin resets the data, its last defined value may be persisted forever in the shot files and in settings.tdb when the user changes the skin or disables the plugin. Base code may also be affected by these problems, though, as happened with the introduction of the "beverage_type" field.
- **Automate many coding tasks**. Creating code that reads the data dictionary and buils a GUI for them, or adds database fields dynamically, makes it easier to build some powerful features. This is extensivelly used by the SDB and DYE plugins, and would be a prerequisite to incorporate SDB functionallity in the base app.
- **Make fields self-documenting**

This implements the proposed API and is only missing validation functionallity.
Example metadata for shot description fields (as used by the DYE extension) are included in app_metadata.tcl as an example. If accepted, new fields will be added in successive versions to cover app needs.

### DIFF
--- a/de1plus/app_metadata.tcl
+++ b/de1plus/app_metadata.tcl
@@ -1,0 +1,318 @@
+
+proc init_app_metadata {} {
+	# Define the properties to be used in the app data dictionary
+	metadata dictionary add domain [list validate_category {shot profile}]
+	metadata dictionary add category validate_nonempty_string
+	metadata dictionary add section validate_nonempty_string
+	metadata dictionary add subsection validate_nonempty_string
+	metadata dictionary add owner_type [list validate_category {base skin plugin}]
+	metadata dictionary add owner
+	metadata dictionary add name validate_nonempty_string
+	metadata dictionary add name_plural validate_nonempty_string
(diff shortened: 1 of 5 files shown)


Mimoja (approved):
Look good to me. @decentjohn ?

ebengoechea:
@decentjohn this needs your approval (friendly reminder ;)

PR #167, opened 2021-06-08
Make limits&parameters for numeric fields in scentone.tcl come from metadata

ebengoechea:
This replaces the hardcoded values of slider options -from, -to, -bigincrement and -resolution of numeric settings variables in Insight scentone.tcl by the equivalent values looked up in the new metadata dictionary.

With this change, option values are defined in a single place instead of hardcoded all around the code, which facilitates maintenance. DYE also looks up those values in the metadata dictionary.

The settings items affected are `espresso_enjoyment`, `grinder_dose_weight`, `drink_weight`, `drink_tds` and `drink_ey`.

As a side effect, this increases the upper limit of `settings(drink_weight)` from 100 to 500 g. This corrects 2 recently reported bugs:
- Pourover actual final weight could not be logged as it is normally above 100 g ([reported by Mitch Hale](https://3.basecamp.com/3671212/buckets/7351439/messages/3762831315#__recording_3837727303))
- Long espresso shots with a final weight above 100 g would show 100 g on DSx home page instead of the actual scale weight ([reported by JasonC](https://decentforum.com/t/100g-bug/1526)), CC @Damian-AU

### DIFF
--- a/de1plus/skins/Insight/scentone.tcl
+++ b/de1plus/skins/Insight/scentone.tcl
@@ -115,8 +115,9 @@ set slider_trough_color2 #f3f3f3
 
 
 # from http://www.iconarchive.com/show/shiny-smiley-icons-by-iconicon.html
+lassign [metadata get espresso_enjoyment n_decimals min max bigincrement] n_dec min max biginc
 add_de1_text "describe_espresso" 80 360 -text [translate "Enjoyment"] -font Helv_8_bold -fill "#7f879a" -anchor "nw" -width 800 -justify "left"
-	add_de1_widget "describe_espresso" scale 300 450 {} -to 100 -from 0 -background #e4d1c1 -showvalue 0 -borderwidth 1 -bigincrement 10 -resolution 1 -length [rescale_x_skin 1000]  -width [rescale_y_skin 150] -variable ::settings(espresso_enjoyment) -font Helv_15_bold -sliderlength [rescale_x_skin 125] -relief flat -command {} -foreground #FFFFFF -troughcolor $slider_trough_color1 -borderwidth 0  -highlightthickness 0 -orient horizontal 
+	add_de1_widget "describe_espresso" scale 300 450 {} -to $max -from $min -background #e4d1c1 -showvalue 0 -borderwidth 1 -bigincrement $biginc -resolution 1 -length [rescale_x_skin 1000]  -width [rescale_y_skin 150] -variable ::settings(espresso_enjoyment) -font Helv_15_bold -sliderlength [rescale_x_skin 125] -relief flat -command {} -foreground #FFFFFF -troughcolor $slider_trough_color1 -borderwidth 0  -highlightthickness 0 -orient horizontal 
 	add_de1_variable "describe_espresso" 1300 600 -text "" -font Helv_8 -fill "#4e85f4" -anchor "ne" -width 600 -justify "left" -textvariable {$::settings(espresso_enjoyment)}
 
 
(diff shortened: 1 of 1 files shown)


decentjohn:
Good stuff.  I've talked to Johanna about moving the ScentOne stuff into an optional extension, as almost nobody uses it, and ScentOne has not done well in the marketplace.  

My thinking was to make the entire ❤️ section in a multi-tab "load your extension gui here" so that REALLY useful things like the HISTORY VIEWER could be easily accessed, and unused things are loaded out.

ebengoechea:
@decentjohn I remember that, but I don't quite get the use case for the multi-tab page for extensions. The tab structure is limiting and may generate conflicts between extensions (what if 2 extensions use the same tab space?). I would definitively **not** want to see the new history viewer there.

The only advantage I see is having a point of entry from Insight's home page, but i too think that's quite problematic. When introducing DYE it was clear that many users, even after months of using Insight, weren't aware that there was "Describe espresso" functionallity "behind" the heart icon. They clicked there once, saw the godshots thing, which they never found a use for, and never looked back again at the other tabs. The same would happen if you "hide" any new functionallity behind the heart (or whatever other) icon. It's not a great UI choice.

decentjohn:
My idea on the ❤️ section was that the 4 tabs would be any-use and an extension would basically define itself as needing some sort of tab.

What I'm trying to recreate is what I see in other apps that have extensions, which is some sort of "blank area" that they can work in. We did this for the the Extension Configure page already.

However, I get your point that tabs are not ideal.

The Chrome approach is not bad at all, of namely a "configure" button that then launches whatever the extension wants, full screen.

I've been thinking about how I might extend advanced profiles with code (ie, "tea type" sets the temp), and I've been trying to decide between a new "config" tab in Advanced, or just a "config" button that then launches its own full screen page.  Both have their advantages.  

I am not at all settled in opinion on this.

decentjohn:
I think that I've previously gone with the reality that Insight, and the tabbed look, was the de1app look, and Settings followed that.

Since DSX and other skins have their own look, this isn't the case any more, so probably giving the skins more flexibility is better.

I don't know if it's possible to let a skin define a "look" so that (for instance) the "choose the tea type" full screen config for a profile, would look different if you were running Insight vs DSX skin.

ebengoechea:
@decentjohn Of course, that's what all DUI theming & styling is about. Both DYE and the new History Viewer modify their aspect depending on the active theme/skin. Each skin needs to provide its own set of styles definitions to get this functionallity, and it is already available for default/Insight, DSx (from DYE at the moment, but will be migrated to DSx itself soon) and MimojaCafe. The limit of this is that it only applies to general options that can be modified at page setup time, like colors, shape of buttons, etc. You can't have background images or icon images (well, you can, but you would have to provide a version of the image for each theme, which at the end is not very operative).

Then there are generic pages like `dui_item_selector` that allow selecting something from a list. Right now it always has the active skin/theme aspect, but I programmed it so it can adapt to different themes in the same session depending on where it's open from: if it's a settings page, it could use the default/Insight theme, but if it's a DSx page, use the DSX theme. This requires using page namespaces and is done by totally deleting the page and its items, then recreating it under the new theme by re-running its `setup` command.

decentjohn:
Oh cool! Is there any sample code I could use to start my adventure with this, specifically so a custom per-profile gui that I have in mind?

ebengoechea:
@decentjohn I would look at the code for generating the page ::dui::pagess::dui_item_selector, line 7838+ in dui.tcl

decentjohn:
Thanks.

PR #168, opened 2021-06-08
Make settings(drink_weight) store espresso weight only (not hot water)

ebengoechea:
**Motivation**
Commit b619ebf5 extended the usage of **$::settings(drink_weight)** to store the weight of the last espresso shot **or the last hot water shot**. Previously that field was used only for espresso shots.

The problem is that `$::settings(drink_weight)` was also being used forshot metadata. End users could modify `$::settings(drink_weight)` after the shot, either in Insight ("My espresso" tab in the Godshots page) or using the DYE plugin. The field was also being stored in the SDB database, and in visualizer.coffee. 

After the change, if a user pours hot water after making a shot, the weight of the hot water is saved on `$::settings(drink_weight)`, overwritting the weight of the last espresso. If now the user tries to edit his last espresso shot, re-upload it to Visualizer, etc., he gets the wrong weight value in the shot file, in the SDB database, and/or in Visualizer.

`$::settings(drink_weight)` is also being used to show the weight of the last espresso shot (and not hot water) in the DSx home page, so the wrong value is shown if hot water has been poured after the espresso shot, as in this capture (the chart shows the espresso drink weight was around 36 g, but the value on the bottom shows 16.8 g instead, the weight of the hot water done after the shot), CC @Damian-AU : 
image

This bug was [reported by JoeD](https://decentforum.com/t/visualizer-upload-bug/1491/3).

**Changes**
A new settings variable is defined that contains the running weight during the shot of either espresso or hot water, so that `$::settings(drink_weight)` is now only used for the after-shot weight of espressos. The new variable is **$::settings(running_weight)**, and its value can be retrieved as a string using command `drink_weight_text` (note that this proc has not changed its name at the moment as it is used in several skins).

**$::settings(drink_weight)** recovers its original meaning and is set only when an espresso shot is finished.

### DIFF
--- a/de1plus/vars.tcl
+++ b/de1plus/vars.tcl
@@ -1104,7 +1104,7 @@ proc drink_weight_text {} {
 		return ""
 	}
 
-	return [return_weight_measurement [ifexists ::settings(running_weight) 0.0]]
+	return [return_weight_measurement $::settings(running_weight)]
 }
 
 proc dump_stack {args} {
--- a/de1plus/vars.tcl
+++ b/de1plus/vars.tcl
@@ -1104,7 +1104,7 @@ proc drink_weight_text {} {
 		return ""
 	}
 
-	return [return_weight_measurement $::settings(drink_weight)]
+	return [return_weight_measurement [ifexists ::settings(running_weight) 0.0]]
 }
 
 proc dump_stack {args} {
--- a/de1plus/device_scale.tcl
+++ b/de1plus/device_scale.tcl
@@ -1002,9 +1002,8 @@ namespace eval ::device::scale::history {
 		}
 
 		set $::device::scale::history::_final_weight_name $cwe
-		set ::settings(drink_weight) [round_to_one_digits $cwe]
+		set ::settings(running_weight) [round_to_one_digits $cwe]
 	}
-
 
 
 
@@ -1436,7 +1435,11 @@ namespace eval ::device::scale::callbacks {
 		}
 	}
 
-
+	proc save_drink_weight {event_dict} {
+		if { [dict get $event_dict previous_state] eq "Espresso" } {
+			set ::settings(drink_weight) [round_to_one_digits $::de1(final_espresso_weight)]
+		}
+	}
 
 	::de1::event::listener::on_major_state_change_add -noidle ::device::scale::callbacks::on_major_state_change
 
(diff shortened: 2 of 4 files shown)


secabeen:
Wouldn't it make more sense to maintain variables for each type of weight-based operation (espresso_weight, hotwater_weight), and then have the generic variable (drink_weight) hold the weight of the most recent operation?  The drink/running split proposed here creates cryptic knowledge (drink is for espresso, and running is for hot water) that is not obvious.

If we want to not break old skins, we can deprecate the drink_weight name wherever it's used, make it an alias to espresso_weight, and come up with some other name for the weight of the most recent drink.

ebengoechea:
@secabeen `$::settings(drink_weight)` was used as the metadata variable for the final espresso weight in the app way before the mentioned commit, which broke features in several skins and extensions (DSx, Insight, SDB, DYE and Visualizer). The intention of the change in the commit (as I've discussed with its author) was to keep the running weight for any of espresso or hot water in a variable, so the change I'm proposing keeps the spirit. 

The change should not have reused a variable that already had a meaning (and was documented so in the DYE manual, the only place that has documentation for shot metadata). In fact I think the variable was introduced originally preciselly for espresso metadata (for the "This espresso" tab in the Godshots/Scentone/Describe espresso Insight page), and its naming is consistent with that (there are also similar `$::settings(drink_tds)` and `$::settings(drink_ey)` variables). 

There are variables for each type of weight-based operation: `$::de1(final_espresso_weight)` and `$::de1(final_water_weight)`. 

I don't care which name is used for a generic variable that keeps the weight of the current weight-based operation, and in fact `running_weight` was proposed by Jeff himself. But changes should either be backwards-compatible or, if not, have very very good reasons for the change, which I don't see in this case.

de1plus/vars.tcl
```diff
@@ -1104,7 +1104,7 @@ proc drink_weight_text {} {
 		return ""
 	}
 
-	return [return_weight_measurement $::settings(drink_weight)]
+	return [return_weight_measurement [ifexists ::settings(running_weight) 0.0]]
```
jeffsf, line 1107: Would it be better to initialize `::settings(running_weight)` in `machine.tcl` so that a run-time check does't need to be made on every call?

ebengoechea: > Would it be better to initialize `::settings(running_weight)` in `machine.tcl` so that a run-time check doesn't need to be made on every call?

You're right. Changing it.

ebengoechea:
Added an additional commit 8a7ae36 to do as @jeffsf suggests (initialize `settings(running_weight)` in machine.tcl and thus remove the runtime check of existence).

decentjohn:
Thanks for doing this, as the various "weight" variables evolved as the functionality of the DE1 improved, leaving some old variables around as functionality was improved.

PR #175, opened 2021-06-26
New wake up scheduler pluging

Damian-AU:
New wake up scheduler pluging


Mimoja:
Did not yet have the time to test, might need a day or two, thank you so much !

Damian-AU:
Could I get feedback on this?  feeling discourage by the silence, please advise if that is the intention?

Mimoja:
@Damian-AU Sorry, this was merged 6 days ago by me and i forgot to push

Mimoja (commented):
the .DS_Store files should not be commited.

de1plus/plugins/DPx_Scheduler/plugin.tcl
```diff
@@ -0,0 +1,356 @@
...
+        set DPx_shed {}
+        foreach k [lsort -dictionary [array names item]] {
+            set v $item($k)
+            append DPx_shed [subst {[list $k] [list $v]\n}]
+        }
+        write_file $fn $DPx_shed
+    }
+
+    proc minutes_digits {} {
+        if {$::plugins::DPx_Scheduler::minutes >= 50 && $::plugins::DPx_Scheduler::minutes < 60} {
```
Mimoja, line 336: You can use "variable" here, which shortens these a bit

like this:
https://github.com/decentespresso/de1app/blob/main/de1plus/plugins/visualizer_upload/plugin.tcl#L231

or this
https://github.com/decentespresso/de1app/blob/4b6a2959032aca7962ac2b7565c71389d4a8a4fc/de1plus/history_viewer.tcl#L77

Mimoja (commented):
Rest looked good

ebengoechea:
I join the suggestion. Using "variable" on namespace procs avoids having to type fully qualified names, you then treat the variables as local variables, plus it is slightly more efficient.

I would also suggest to avoid introducing any new global variables in an extension. Under normal circunstances all its data should be kept in namespace variables. You already removed several globals, but still have ::DPx_scheduler_times and ::DPx_scheduler_minutes.

PR #184, opened 2021-07-18
Make DUI dscales work on tablets

ebengoechea:
This PR completes the DUI version of scales, called "dscales". They offer an alternative to Tk scale widgets, built from canvas primitives. Previous versions of "dscale" worked on PC but failed on android. Now they should work on all platforms, plus several improvements have been made to make the `-resolution` option work correctly for setting discrete values.

A new command `dui platform button_motion` has been introduced, that, similar to analougosly named commands, returns the virtual event corresponding to button or finger motions depending on the current platform.

To reviewers: there is no code currently in the app that shows a "dscale", so it can be tested. To test it, I suggest to manually replace line  1102 in file `skins/Insight/skin.tcl`:
```tcl
add_de1_widget "steam steam_1 steam_3" scale 10 1436 {} -from 40 -to 250 -background "#e8e1df" -borderwidth 1 -showvalue 0  -bigincrement 100 -resolution 10 -length [rescale_x_skin 2000] -width [rescale_y_skin 150] -variable ::settings(steam_flow) -font Helv_10_bold -sliderlength [rescale_x_skin 500] -relief flat -command {set_steam_flow} -orient horizontal -foreground #FFFFFF -troughcolor "#d7d9e5" -borderwidth 0  -highlightthickness 0
```
by this one:
```tcl
dui add dscale "steam steam_1 steam_3" 40 1510 {} -from 40 -to 250 -bigincrement 100 -smallincrement 10 -resolution 10 -length 1950 -width 14 -sliderlength 120 -variable ::settings(steam_flow) -command {set_steam_flow} -orient horizontal
```

which produces this:
image
﻿

PR #185, opened 2021-07-19
Changes the steam max flow rate scale widget to a DUI dscale

ebengoechea:
Changes the steam max flow rate scale widget in Insight steam page from a Tk scale to a DUI dscale, as [requested](https://github.com/decentespresso/de1app/pull/184#issuecomment-882183053) by @decentjohn.

### DIFF
--- a/de1plus/skins/Insight/skin.tcl
+++ b/de1plus/skins/Insight/skin.tcl
@@ -1100,8 +1100,8 @@ add_de1_text "steam" 1840 250 -justify right -anchor "nw" -text [translate "Info
 
 
 	# realtime control over the steam flow rate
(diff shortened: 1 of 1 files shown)


decentjohn:
I'm unsure if the behavior of a fast double-tap on the +/- causing a BIG jump, makes sense.  Let's bring that up on Diaspora for discussion. Seems to me that if people put the trouble to go to +/- then multiple taps would cause multiple small incremental changes.  A Big change is easy to effect by placing your finger at the new position.

marcello3d:
I find the original double tap behavior mildly annoying. It never triggers when I want and always when I don't. :)

ebengoechea:
> I'm unsure if the behavior of a fast double-tap on the +/- causing a BIG jump, makes sense. Let's bring that up on Diaspora for discussion. Seems to me that if people put the trouble to go to +/- then multiple taps would cause multiple small incremental changes. A Big change is easy to effect by placing your finger at the new position.

Easy to remove or change. Please keep the feedback coming and I'll make the desired changes at once.

PR #188, opened 2021-08-03
Added plugin for old_lcd_disable

bkeller2:
Pretty simple plugin to restore the previous LCD scale functionality that were changed in 64ab57cca3aa1111e429b02cb18b5eaa096603f9. 

It was tested both on desktop (macOS) and on a tablet and seemed to have no issues.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -27,3 +27,6 @@
 [submodule "de1plus/plugins/DPx_Screen_Saver"]
 	path = de1plus/plugins/DPx_Screen_Saver
 	url = https://github.com/Damian-AU/DPx_Screen_Saver
+[submodule "de1plus/plugins/old_lcd_disable"]
+	path = de1plus/plugins/old_lcd_disable
+	url = https://github.com/bkeller2/old_lcd_disable


Mimoja:
Looks good, thank you. Pulling into nightly now

Mimoja:
As a Nitpick (that we are absolutely not enforcing...) in the "industry" a commit message is often worded imperatively:

https://chris.beams.io/posts/git-commit/

So this commit would have been:
"Add plugin for old_lcd_disable"

You might disagree with this style, just wanted to leave you proper feedback :)

bkeller2:
Useful feedback thanks!

decentjohn:
I've no problem at all with this plugin being here, but do want you to know that my intention is to restore the old "LED turns off" functionality in the future, for scales that are plugged into USB power.  Battery powered scales will keep the current LCD-stays-on-during sleep.

bkeller2:
@decentjohn Will this be the case for Skale2? I didn't think there was an option to determine that via skale2's protocol.

decentjohn:
@bkeller2 it's in the skale API:

binary scan $value cucucucucucu t0 t1 
msg "- battery level: $t0/$t1"

https://decentespresso.com/scale_api

PR #190, opened 2021-09-14
add settings entry to disable Decent Scale LCD on sleep

silasg:
As far as I understood the discussion between @decentjohn and @ebengoechea in Diaspora, a setting to re-enable disabling the Decent Scale LCD when the machine goes to sleep would be acceptable. Since I really miss this tiny change, here's a PR for it.

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -35,10 +35,9 @@ proc scale_disable_lcd {} {
 		}
 	} elseif {$::settings(scale_type) == "decentscale"} {
 		
-		set do_this 0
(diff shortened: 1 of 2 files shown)


Mimoja:
Please use the old_lcd_disable plugin instead

Mimoja:
https://github.com/bkeller2/old_lcd_disable/blob/main/plugin.tcl

silasg:
Ah thanks, wasn't aware of the plugin. Maybe you could delete the dead if in bluetooth.tcl and/or update the comment to point to the plugin ;)

PR #212, opened 2021-11-11
Makes screensaver file lookup case insensitive

nowstuseeit:
The screensaver doesn't read files that are not *.jpg (such as *.JPG), this makes the filetype in the lookup case insensitive.

### DIFF
--- a/de1plus/utils.tcl
+++ b/de1plus/utils.tcl
@@ -679,7 +679,8 @@ proc random_saver_file {} {
             }
         }
 
-        set ::saver_files_cache [glob -nocomplain "[saver_directory]/${::screen_size_width}x${::screen_size_height}/*.jpg"]
+	set saver_path "[saver_directory]/${::screen_size_width}x${::screen_size_height}/"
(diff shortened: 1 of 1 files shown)


Mimoja:
Hi and thank you for your first PR. Looks good to me and could be merged. 

I think this could also be solved with a lsearch on the resulting glob.
```
lsearch -nocase -all -inline -glob [glob ./*] {*.jpg}
```
I would leave it to you which ever version you prefer, i think both a equally easy / hard to read. 

Is this tested code? Tested just on desktop? Any issues I should be looking for before merging?

@johnbuckman should we also support png images in the near future? DUI already does.  While i see the appeal of them I think the additional size when downloading really does not outweight the quality benefits (tho JPEGs can be made huge as well, so what gives?)

CC: @ebengoechea

nowstuseeit:
No, it is not tested code outside of running on my machine (on stable build v1.37.0) where it works as intended. The issue isn't evident on desktop (windows at least). I assume the emulator ignores case in the file system. With the changed utils.tcl file the desktop (windows, stable and nightly 1.38.4.1) app continues to work fine.

I don't mind which way it is solved, this is probably more of a bug report with a proposed solution.

PR #213, opened 2021-11-14
Convert individual shots to v2 & profile basic to advanced with any array

ebengoechea:
This refactors a couple of shot & profile commands to work on arbitrary files instead of being bound to the global settings or all shots at once. This is in preparation of the forthcoming version of DYE that allows users to export past shot files and to view profiles from past shots.

- New command `shot::convert_legacy_to_v2` converts just one file. Code is taken from `shot::convert_all_legacy_to_v2`, which has been refactored to use the new command on each history file, to avoid code duplicity.

- `profile::pressure_to_advanced_list`, `profile::flow_to_advanced_list`, and `profile::settings_to_advanced_list` get an optional argument `settingsvar` that specifies the global variable they should use. This allows to make the conversion of basic (flow/pressure) profiles to advanced profiles not only on the global settings but on any arbitrary array having profile data (such as saved profiles or past shots read from history).

### DIFF
--- a/de1plus/profile.tcl
+++ b/de1plus/profile.tcl
@@ -488,4 +488,30 @@ namespace eval ::profile {
             }
         }
     }
+    
+    # "sanitize" filename from profile title. Refactored from proc save_profile, so this is reusable in other parts
+    # of the code such as profiles importing from Visualizer.
+    proc filename_from_title { title } {
+        set fn $title
+        regsub -all {\s+} $fn _ fn 
+        regsub -all {\/+} $fn __ fn 
+        regsub -all {[\u0000-\u001f;:?<>(){}\[\]\|!@#$%^&*-\+=~`,.'"]+} $fn "" fn
+        return [string range $fn 0 59]
+    }
+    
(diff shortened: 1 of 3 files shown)


ebengoechea:
Added commit 17c1af1 which adds 2 new profile utility commands: `profile::filename_from_title` & `profile::fix_profile_type`.

These just copy existing code in procs `select_profile` and `save_profile`, so they are now resusable in different situations. For example, the new commands are used by DYE to import profiles from past shots or from Visualizer, which couldn't use the existing commands as those commands are bound to the global settings.

 `select_profile` and  `save_profile` have been refactored to invoke the new commands, so as to avoid code duplication.

ebengoechea:
Added commit b078947 that fixes a bug in both `shot::convert_all_legacy_to_v2` and 
`shot::convert_legacy_to_v2` (which copied code from the former) without being restored, which meant the latest user
settings could be lost when these were invoked, as they were being replaced by the settings available when the converted shot was done, and were not replaced back.

PR #216, opened 2021-11-18
Profile read and textual

ebengoechea:
This adds new features to the `::profile` namespace and modifies a few things to make them more generic, so they can be used in the next version of the DYE Profile Viewer:

# New
- New command `profile::read_legacy`, reads a legacy (.tcl) profile into an array. This is needed because the standard profile loading command is `load_settings_var`, which loads directly into the global `$::settings` variable, whereas the new command will restrict the read variables to those from profiles, and return the result as a list that can be set to an array.
- New command `profile::legacy_to_textual` transforms a legacy (.tcl) profile into a dictionary with a human-readable textual representation of the profile. This has been refactored from code in the DYE plugin so it can be easily reused by all skin & plugin programmers.

# Changed
- The 3 commands `profile::pressure_to_advanced_list`, `profile::flow_to_advanced_list`  and `profile::settings_to_advanced_list` now use `upvar` on its first optional argument, and thus can be applied to any list, including local variables in the invoking proc. Still default to use `::settings` if  they receive no argument, so totally backwards-compatible. 
- Step names when generating `advanced_shot` from basic flow or pressure profiles are not translated anymore. Translation should be done only in the GUI, and this could produce translated steps in the profile file, which if are ever translated into a third language because the user changed his app language, would never be translated back.

### DIFF
--- a/de1plus/profile.tcl
+++ b/de1plus/profile.tcl
@@ -513,5 +513,288 @@ namespace eval ::profile {
         
         return $profile_type
     }
+        
+    # Based on proc load_settings_vars, but only loads profile variables, and returns a list with the profile (which can be
+    # coerced to an array) instead of loading the profile into the global settings. The source file can also be a shot 
+    # file, and only the profile-related vars will be loaded from it, plus any variables given in list $extra_vars.
+    proc read_legacy { profile_path {extra_vars {}} } {
+        if { [file pathtype $profile_path] eq "relative" && [file dirname $profile_path] eq "." } {
+            set profile_path "[homedir]/profiles/[file tail $profile_path]"
+        }
+        if { [file extension $profile_path] eq "" } {
+            append profile_path ".tcl"
+        }
+        if { ![file exists $profile_path] } {
+            msg -WARNING [namespace current] load_profile: "profile file '$profile_path' not found"
+            return
+        }
+        
+        msg -INFO [namespace current] read_profile: "reading file '$profile_path'"
+        array set profile {}
     
+        # Set defaults in case they are not defined in the profile file 
+        set profile(final_desired_shot_volume_advanced_count_start) 0
+        set profile(settings_profile_type) "settings_2a"
+        set profile(beverage_type) "espresso"
+    
+        try {
+            array set arrfile [encoding convertfrom utf-8 [read_binary_file $profile_path]]
+        } on error err {
+            msg -WARNING [namespace current] read_profile: "error reading profile '$fn': $err"
+            return
+        }
+    
+        foreach k [concat [::profile_vars] $extra_vars] {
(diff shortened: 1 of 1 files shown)


de1plus/profile.tcl
```diff
@@ -513,5 +519,282 @@ namespace eval ::profile {
...
+            set profile(final_desired_shot_volume) $profile(final_desired_shot_weight)
+        }
+    
+        if { $profile(advanced_shot) eq {} } {
+            # Ensure the profile's advanced_shot variable is always defined
+            switch $profile(settings_profile_type) \
+                settings_2a {
+                    array set temp_profile [pressure_to_advanced_list profile]
+                    set profile(advanced_shot) $temp_profile(advanced_shot)
+                } settings_2b {
```
Mimoja, line 583: I dont think they need to be in the same line like if / else

de1plus/profile.tcl
```diff
@@ -513,5 +519,282 @@ namespace eval ::profile {
...
+                    array set temp_profile [flow_to_advanced_list profile]
+                    set profile(advanced_shot) $temp_profile(advanced_shot)
+                }
+        }
+        
+        return [array get profile]
+    }
+    
+    # Returns a dictionary with a textual representation of the provided profile.
```
Mimoja, line 592: would it make sense to convert to json-compat advanced list first and then work on that? That way we at least dont add dependencies onto the TCL profiles

Mimoja:
rest looks absolutely good to me

de1plus/profile.tcl
ebengoechea: I thought JSON versions were mainly for exporting, not for internal processing, or that was only for shots? Shouldn't be too hard to change, but I would leave that for later.

de1plus/profile.tcl
ebengoechea: Do you mean that the `switch` line would work without the final "\"?

de1plus/profile.tcl
Mimoja: I would prefer to remove the TCL profiles in the future, but as long as we are forced to deal with GUI elements tied to ::settings entries it will most likely not be a realistic goal to get the migration up and running smoothly.

de1plus/profile.tcl
Mimoja: settings_2a {
    } 
   settings_2b {

ebengoechea: Ah, ok, sure can be changed, do it on merge if you want

de1plus/profile.tcl
ebengoechea: Ok, sorry I didn't realize to do it directly with the new vesion. If needed I'll be happy to migrate it.

PR #218, opened 2021-11-24
Legacy profile management commands

ebengoechea:
New and modified commands in the `::profile` namespace, continuing with the objective of providing a set of reusable commands to manage profiles that work on input and output data different of the `settings` global variable. All of this is needed for the forthcoming version of DYE.

- New command `profile::profile_type_text` returns an english textual description of the profile type that includes both `settings_profile_type` and `beverage_type` in one.
- New command `profile::find_file` returns the full path for an existing profile filename (both legacy and v2)
- New command `profile::modify_legacy` to easily modify parts of legacy profile files.
- New command `profile::import_legacy`, taken from DYE, handles importing legacy profiles from sources such as history shots or Visualizer downloads.

Also, the existing command `profile::read_legacy` is now more versatile, allowing different types of input data.

### DIFF
--- a/de1plus/profile.tcl
+++ b/de1plus/profile.tcl
@@ -520,6 +520,32 @@ namespace eval ::profile {
         return $profile_type
     }
         
+	proc profile_type_text { profile_type {beverage_type {}} } {
+		switch [fix_profile_type $profile_type] \
+		settings_2a {
+			set type "Pressure"
+		} settings_2b {
+			set type "Flow"
+		} settings_2c {
+			set type "Advanced"
+		} default {
+			set type "Unknown"
+		}
+		
+		switch [lindex $beverage_type 0] \
+			pourover {
+				set beverage_type "pour-over "
+			} tea_portafilter {
+				set beverage_type "tea "
+			} {} {
+				set beverage_type ""
+			} default {
+				set beverage_type "$beverage_type "
+			}
+		
+		return "${type} ${beverage_type}profile"
+	}
+	
     # Based on proc load_settings_vars, but only loads profile variables, and returns a list with the profile (which can be
     # coerced to an array) instead of loading the profile into the global settings. The source file can also be a shot 
     # file, and only the profile-related vars will be loaded from it, plus any variables given in list $extra_vars.
--- a/de1plus/profile.tcl
+++ b/de1plus/profile.tcl
@@ -520,32 +520,59 @@ namespace eval ::profile {
         return $profile_type
     }
         
-	proc profile_type_text { profile_type {beverage_type {}} } {
-		switch [fix_profile_type $profile_type] \
-		settings_2a {
-			set type "Pressure"
-		} settings_2b {
-			set type "Flow"
-		} settings_2c {
-			set type "Advanced"
-		} default {
-			set type "Unknown"
-		}
-		
-		switch [lindex $beverage_type 0] \
-			pourover {
-				set beverage_type "pour-over "
-			} tea_portafilter {
-				set beverage_type "tea "
-			} {} {
-				set beverage_type ""
-			} default {
-				set beverage_type "$beverage_type "
-			}
-		
-		return "${type} ${beverage_type}profile"
-	}
-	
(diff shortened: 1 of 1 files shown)


de1plus/profile.tcl
```diff
@@ -520,41 +520,110 @@ namespace eval ::profile {
...
+            set type "Flow"
+        } settings_2c {
+            set type "Advanced"
+        } default {
+            set type "Unknown"
         }
-        if { [file extension $profile_path] eq "" } {
-            append profile_path ".tcl"
+        
+        switch [lindex $beverage_type 0] \
```
Mimoja, line 535: dont we maybe want to strip the string here first?

de1plus/profile.tcl
```diff
@@ -520,41 +520,110 @@ namespace eval ::profile {
...
-        if { ![file exists $profile_path] } {
-            msg -WARNING [namespace current] load_profile: "profile file '$profile_path' not found"
-            return
+
+        if { [file extension $filename] eq "" || [file extension $filename] ne $ext } {
+            append filename $ext
         }
         
-        msg -INFO [namespace current] read_profile: "reading file '$profile_path'"
+        if { [file pathtype $filename] eq "relative" && [file dirname $filename] eq "." } {
```
Mimoja, line 562: shouldn't we handle ./ and ./profiles as well (at least in theory)?

de1plus/profile.tcl
```diff
@@ -520,41 +520,110 @@ namespace eval ::profile {
...
+        } else {
+            set profile_vars [concat [::profile_vars] $extra_vars]
         }
-    
-        foreach k [concat [::profile_vars] $extra_vars] {
-            if { [info exists arrfile($k)] } {
-                set profile($k) $arrfile($k)
-            } elseif { ![info exists profile($k)] } {
-                set profile($k) {}
+
```
Mimoja, line 621: nit: whitespace

Mimoja (approved):

de1plus/profile.tcl
ebengoechea: "./" is handled, because both `[file dirname "./abc.tcl"]` and `[file dirname "abc.tcl"]` return ".".

"./profiles" is not handled, true. I didn't expect that type of input normally, only either filenames or full paths.

de1plus/profile.tcl
ebengoechea: Thing is I found some profiles having `beverage_type {pourover}` instead of `beverage_type pourover`, so I wanted to handle those.

de1plus/profile.tcl
Mimoja: okay, can we add a warning then? if is relative && ![file dirname $filename] eq "."

de1plus/profile.tcl
Mimoja: meant something like:
switch  [ string trim [lindex $beverage_type 0] " " ]

de1plus/profile.tcl
ebengoechea: Not sure what's the whitespace issue here , is it the blank line 621? I see many lines with that (variable number of whitespace characters), would you like those removed on all blank lines or what?

de1plus/profile.tcl
ebengoechea: But I think the parsing of the profile file should prevent `beverage_type` to contain spaces.

de1plus/profile.tcl
ebengoechea: But you will already get the warning that the file does not exists: "de1plus/profiles/profiles/default.tcl"

de1plus/profile.tcl
ebengoechea: Doesn't hurt, in any case. Changed.

de1plus/profile.tcl
ebengoechea: Added case to handle "./profiles" in a new commit. Plus, spotted a bug with a hardcoded `profiles` instead of `$folder`.

PR #223, opened 2021-11-30
Use Insight graph series colors in history viewer default DUI theme

ebengoechea:
This modifies the colors of the graph series lines in the default history viewer DUI theme, to match those used in the Insight/default skins. Skins wanting to use other colors can redefine them in their own themes.


ebengoechea:
There is a link to it (top-right) on DYE and I know some people use it despite its current shortcomings. But mainly I'm changing it because I'm reusing the styles for the graph in the new Shot Selector preview graph, which I'll PR now.

PR #263, opened 2022-09-10
Acaia Lunar 2021 Fix [possibly others]

xander-io:
Full write-up can be found [here](https://public.3.basecamp.com/p/opKQmZ9CJRorLkPU41wpbqwu)

Features:
- Increased tickle timeout to ensure we're actually disconnected before displaying to the user
- Completely revamped how we parse data on acaia:
    - Scan the buffer to see if any data is viable.
    - Throw away messages that don't have the correct msg or event type (fixes watchdog misfire)
    - Added an arbitrary length threshold in case we do find a valid msg but the data is too long to care about (also fixes watchdog misfire/blips in weight graph)
    - Added support for event type 11 to give us a greater chance of having valid data


Mimoja:
Looks fine to me. Did not immediately crash, testing in nightly now :)

PR #264, opened 2022-09-25
Filter coffee Paperless profile

beettlle:
This is a modification of Scott Rao's 2.1 profile that doesn't use a paper filter.


decentjohn:
Before we add a new profile to the app, it needs to demonstrate itself as being useful to the community.  Otherwise, we'd have hundreds of profiles that people have created.

Have you uploaded your profile to Diaspora, described how it works, gotten others to try it out?

beettlle:
I 100% Agree @decentjohn .  I put this up as I was already looking around the code to see if I could fix [this issue](https://github.com/decentespresso/de1app/issues/265). 
I started a thread at the same time as I posted this PR over [here](https://3.basecamp.com/3671212/buckets/7351439/messages/5364440874).

decentjohn:
So, generally, we only put profiles into the app itself, that rise to at least 1% of use as reported by https://visualizer.coffee/

beettlle:
Understood.

PR #268, opened 2023-01-10
Support Acaia Pearl 2021 scales

lgladdy:
Support PEARL 2021 scales under the Acaia Pyxis profile, as well as PEARLS.

Pearl 2021 scales appear with the name `PEARL-{last3hex}`

I manually made this change and copied bluetooth.tcl over to my tablet using File Manager+, then tested with my Pearl 2021 scales.

The scales seem to be working correctly (https://visualizer.coffee/shots/e176dd5c-b744-4ea3-a4fd-1b8b6495213b).

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -1531,7 +1531,8 @@ proc de1_ble_handler { event data } {
  					|| [string first "PROCH" $name]    == 0 } {
 					append_to_peripheral_list $address $name "ble" "scale" "acaiascale"
 					
-				} elseif {[string first "PEARL" $name] == 0 \
+				} elseif {[string first "PEARLS" $name] == 0 \
+ 					|| [string first "PEARL-" $name]   == 0 \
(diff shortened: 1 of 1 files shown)


Mimoja:
Hey, while this is looking good it breaks the original PEAL and PEARL-S Integration. I would therefore suggest to check for "PEARL-" first and then check for "PEARL" second to not break those.  I have never seen an original Pearl appear as "PEARL" only as "PROCH" so it might just be a documentation error on the forums that a firmware with this announcement exists. 

Would you please update your PR to reflect those changes (basically just adding your new line instead of deleting the old one)?

lgladdy:
Hey @Mimoja, Yeh happy to edit this.

My reading of the original line is that it just looked device names starting "PEARLS", so by removing the S as in my PR, my assumption was one match would work for both Pearl 2021s and Pearl S's (ie, matching both "PEARLS-0A0A0A" and "PEARL-0A0A0A".

lgladdy:
@Mimoja Tested and updated!

PR #270, opened 2023-03-21
Add support substate 217, front switch off

Damian-AU:
Add support new FW substate 217, no A/C due to front switch turned off
Show a new page (no_ac) to advise user to turn the front switch on if substate equals 217


decentjohn:
Could you revert this for now? I'm literally about to go from BETA to STABLE and this is a lot of untested changes to put it, including a change standard include, right before going stable.

I'm fine re-merging this change after we go to stable.

The only thing I'm waiting for right now to go to STABLE is finalized translations.

decentjohn:
At the moment, this feature is of limited use, because the firmware won't allow a ble connection until it completes its self test, and the self test can fail if the standby switch is pushed in.  A future firmware version will hopefully allow an earlier ble connection, so the app can indicate the issue.

Mimoja:
Absolutely, worked on my machine and it felt like not too much code so I merged it :)

decentjohn:
I spoke to Ray today about this, and it turns out that I was not in the loop on this.

Ray added firmware support for this new substate 217, which I gathered, but more importantly (for me) Ray also fixed the bug that caused the de1app to not be able to connect to the DE1 during its power up sequence.

This means:
- this new substate 217 feature that @Damian-AU has coded, needs new firmware to work
- the new feature Damian has coded will help people regardless of when they push the standby button, as the de1app can now connect to the DE1 during bootup, when hitting the standby switch has caused a problem
- also, incidentally, the de1app can now have the "tap here to fill the tank" when people have an empty water tank and a refill kit, something which used to be possible but stopped working when the delayed-bluetooth connection issue appeared.

decentjohn:
Ray has another week or two of testing he wants to do before we put the new firmware into Nightly.

This new substate-217 detection can be remerged when the new firmware is available.

PR #272, opened 2023-05-16
Added basic smartchef scale support

randomcoffeesnob:
This allows the usage of the Reflex Smartchef scale. It is available online for around 25$
Unfortunately there is no software-triggered taring. So the button needs to be pressed.

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -119,6 +119,8 @@ proc scale_enable_weight_notifications {} {
 		hiroia_enable_weight_notifications
 	}  elseif {$::settings(scale_type) == "eureka_precisa"} {
 		eureka_precisa_enable_weight_notifications
+	}  elseif {$::settings(scale_type) == "smartchef"} {
+		smartchef_enable_weight_notifications
 	}
 }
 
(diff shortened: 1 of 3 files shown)


de1plus/bluetooth.tcl
```diff
@@ -946,6 +948,45 @@ proc decentscale_tare {} {
...
 }
 
+#### Smartchef Scale
+proc smartchef_enable_weight_notifications {} {
+	if {$::de1(scale_device_handle) == 0 || $::settings(scale_type) != "smartchef"} {
+		return
+	}
+
+	if {[ifexists ::sinstance($::de1(suuid_smartchef))] == ""} {
+		::bt::msg -DEBUG "decent scale not connected, cannot enable weight notifications"
```
baselsalam, line 958: Should this be "Smartchef" scale?

de1plus/bluetooth.tcl
```diff
@@ -946,6 +948,45 @@ proc decentscale_tare {} {
...
+}
+
+proc smartchef_tare {} {
+	#software based taring is not supported by the scale
+	if {$::de1(scale_device_handle) == 0 || $::settings(scale_type) != "smartchef"} {
+		return
+	}
+
+	if {[ifexists ::sinstance($::de1(suuid_smartchef))] == ""} {
+		error "Smartchef Scale not connected, cannot send tare cmd"
```
baselsalam, line 972: Will this show up as a user notification? If so if might be useful to change this to be instructional: Warning: Auto-Tare not available, please hit Tare on the scale"

randomcoffeesnob: Well, I wasn't sure about that. Doesn't that get annoying to see every time? An alternative would be to save the weight at taring and subtract it from the current weight. This way the app would have an "soft" tare. I implemented this behavior in  despresso but didn't dare to do it here. I wanted to make it as simple as possible.

de1plus/bluetooth.tcl
randomcoffeesnob: Yes, of course you're right. Missed that when copying the code

randomcoffeesnob:
Resolved the issues. Thanks for the review.

Mimoja:
Looks good to me, we can merge from my perspecitve. I was unable to get hands on the scale tho, so i can not guarantee function

baselsalam:
FYI! Looks like Tare is supported now?
The video posted on Amazon shows the app has a "Tare" button: https://a.co/d/6GWVwdR

randomcoffeesnob:
Yes, you're right - the app is for several versions and not all support bluetooth taring. For this scale (RX402b20) there is an error message in some cases, that this is not implemented.

randomcoffeesnob:
Can you check the name of the scale with a BLE scanner? There was a discussion in one of the forums that there might be a new version which works differently. Unfortunately we are always a step behind since almost none of the companies really give any information about the devices, how they work etc.

NickGeek:
Thanks, @randomcoffeesnob 

I think it's called "SC02". The model number is RX402b20

NickGeek:
All the details I could see

randomcoffeesnob:
Yes, that's this new scale. We had a discussion concerning Beanconqueror. This version is advertising the name SC02 instead of "smartchef". The model number is the same but it is technically completely different. It also doesn't have the chef picture on the weight plate.

We are trying to figure out how it would need to be implemented, but we don't have a developer who has access to it. And you cannot order the scales to europe.

NickGeek:
Okay, I have access to it but I'm not sure where I'd start for making it work.

NickGeek:
Okay, so the initial connection looks basically the same between `smartchef` and `SC02` (the code is effectively identical).

The difference is in the data basket, the data looks like so:
| constant | weight reading | mode | constant | unknown | constant |
|----|----|---|---|---|--|
| `14FFCA100F0302` | 6 hex digits | `5A0005` (grams) | `0050FB1987E3C` | 3 hex digits | `050953433032`

So `14FFCA100F03020000005A00050050FB1987E3C552050953433032` is 0 grams reading in grams mode and `14FFCA100F03020000115A00050050FB1987E3C543050953433032` is 1.7 grams reading in grams mode (the hex value is 17, you have to divide it by 10)

The division rules for the weight value are as follows:
- `5A0005`: weight / 10
- `5A0002`: weight / 100
- not `5A0000`, `5A0001`, `5A0002`, `5A0003`, `5A0004`, `5A0005`, `5A0008`: weight / 1
- otherwise: weight / 1000

I hope this helps @randomcoffeesnob

Issue #292, opened 2024-01-08
can't change color of the corners of a dbutton


decentjohn:
![Screenshot 2024-01-08 at 7 47 01 PM](https://github.com/decentespresso/de1app/assets/11965250/c82d2a48-fdca-454e-867f-9e20221570f5)

When I change a button's outer shape, such as with:
.can itemconfigure profile_4_btn-out -fill "#ff0000"
.can itemconfigure profile_4_btn-btn -fill "#ff0000"

the corner colors are not affected.

ebengoechea:
I'm aware. Problem is that arcs (used for the corners) and lines (used for the borders) have different arguments in Tk canvas commands. You set straight lines color using ``-fill`` but arc lines color using ``-outline``. That's difficult to solve.

In the meantime, use the 4 corners tags to change the color. I will add a common ``-cor`` suffix tag in DUI to change all at the same time:
```
dui item config $page {profile_4_btn-out-nw profile_4_btn-out-ne profile_4_btn-out-se profile_4_btn-out-sw} -outline "#ff0000"
```

decentjohn:
Given your naming convention of tags (ie out-n and out-s) I had assumed you'd do -ne, -nw, etc... but grepping the DUI code, I couldn't find any use of that label.

And when I try it, there is no effect:
`.can itemconfigure profile_4_btn-btn-ne -fill "#ff0000"`

so perhaps the -ne, -nw, etc...  tag naming of corners never happened in code?

ebengoechea:
I'll look into all pending DUI issues when I'm more advanced with my DSx2-DYE integration 

About ``.can itemconfigure``, I thing it expects item IDs and not tag names.

decentjohn:
It takes either:
`[pathName itemconfigure tagOrId ?option? ?value? ?option value ...?]`(https://www.tcl.tk/man/tcl8.0/TkCmd/canvas.htm#M52)

and I use that fact successfully:
`.can itemconfigure profile_1_btn-btn -fill $b1c`

ebengoechea:
Oh, right, too long without programming Tcl. The error was in using ``-fill`` instead of ``-outline``, which is what changes the color of arcs sections, as I commented above.

Issue #288, opened 2024-01-04
logic error 


decentjohn:
If this first test fails, $is_rethemed is not yet assigned and thus throws a Tcl error.

```
proc retheme { pages new_theme {force 0} } {
	if { ![dui theme exists $new_theme] } {
		msg -ERROR [namespace current] retheme: "new theme '$new_theme' is not a valid theme"
		**return $is_rethemed**
	}
```

Issue #282, opened 2023-12-30
Widgets added with dui::page::add_items may not show


ebengoechea:
Already created widget items added to other pages with ``dui::page::ad_items`` may not show, due to z-order problems (they may be hidden by other canvas elements). This should be fixed on ``dui::page::load`` .

(Will fix this myself, adding the issue for easier reference)

Issue #280, opened 2023-12-30
DUI long-presses bug


ebengoechea:
DUI buttons long-press doesn't work correctly on Android in scenarios with several buttons. Only Android, doesn't happen on PC nor Mac.

Reported by @Damian-AU, currently under investigation.

(I plan to fix this myself, adding the issue for reference)

PR #293, opened 2024-01-12
DUI fixes (z-order, longpress, retheme, dbutton corners)

ebengoechea:
A few minor DUI bug fixes that were pending:

1. Adding items already added to a page to another page using ``dui::page::add_items`` could make the items non-visible due to z-order problems. Fixed. Solves issue #282 

2. A possible fix for the problem of DUI longpresses of dbuttons in android, as described in issue #280. This seems to fix it on my tests but as the problem is hard to reproduce, I need @Damian-AU to check and verify.

3. ``dui::page::retheme`` would raise a runtime error due to an undefined-variable return value if the new_theme didn't exist, as reported by @decentjohn on issue #288, fixed.

4. DUI "round outline" dbutton corners now share a common tag suffix ``<main_tag>-out-cor`` so they can be modified at once. The straight lines in the outline also receive a new common tag suffix ``<main_tag>-out-lin``. See issue #292.

### DIFF
--- a/de1plus/dui.tcl
+++ b/de1plus/dui.tcl
@@ -7705,33 +7705,33 @@ namespace eval ::dui {
 
 			if { $radius1 > 0 } {
 				lappend ids [$can create arc $x0 [expr {$y0+$radius1+1.0}] [expr {$x0+$radius1+1.0}] $y0 -style arc -outline $colour \
-					-width $arc_width -tags [list ${main_tag}-nw {*}$tags] -start 90 -disabledoutline $disabled -state "hidden"]
+					-width $arc_width -tags [list ${main_tag}-nw ${main_tag}-cor {*}$tags] -start 90 -disabledoutline $disabled -state "hidden"]
 			}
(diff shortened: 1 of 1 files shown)


ebengoechea:
@decentjohn Please beware that my pull request seems to have lost your recent line ``set settings(app_title) $::settings(skin)``. Sorry about that, could you please add it back? I thought I had properly rebased just before issuing the request...

decentjohn:
App title was this here:

https://github.com/decentespresso/de1app/commit/057fc2ff01b0effbe0eb890b4b8c46106c16c6f7

do you want to do it, so you're the one messing with dui?  Or I can, whatever works for you.

We can also remove

`proc rectangle {contexts x1 y1 x2 y2 colour } `

now, and I'll put it back into MC.

decentjohn:
It's all working ok for me.  I checked, the app title = skin is there for me.

ebengoechea:
Yep 👍🏻

PR #302, opened 2024-02-07
DUI Fixes

ebengoechea:
This fixes a couple of pending problems in DUI and adds some new features:

- Change the default longpress threshold time from 300 to 1500 ms, as per PulakB recommendation.
- Add new options ``-pressfill``, ``-labelN_pressfill``, and ``-symbolN_pressfill`` to ``dui::add::dbutton``. These allow to change the colors of button backgrounds and labels when a button is pressed. All 3 take a list of colors and integer times in milliseconds for each color change, and at the end always revert back to the original colors. If some times are not provided, will change to each new color (from left to right in the list) every 40 ms. If only one color is given, will change it for 200 ms.
- Allow label and symbol coordinates inside dbuttons (as given by options ``-label_pos`` and ``-symbol_pos``) to accept absolute coordinates. Values between 0.0 and 1.0 are still interpreted as percentage coordinates, but values bigger than 1.0 are assumed to be absolute coordinates from the top-left corner of the button.
- Fix a bug in ``dui::item::dtoggle`` that, when the "widget" was moved at runtime, would show the center circle in the original coordinates.

I'm not automatically merging this as DUI modifications can have unforeseen side effects and I prefer this to be reviewed before merging. I have had these changes for one week on my tablet, and tested all major skins on my PC, and have found no problem.

### DIFF
--- a/de1plus/dui.tcl
+++ b/de1plus/dui.tcl
@@ -8902,8 +8902,18 @@ namespace eval ::dui {
 				}
 				
 				set "label${suffix}_pos" [dui::args::get_option -pos {0.5 0.5} 1 "label${suffix}_args"]
-				set "xlabel$suffix" [expr {$x+int($x1-$x)*[lindex [subst \$label${suffix}_pos] 0]}]
-				set "ylabel$suffix" [expr {$y+int($y1-$y)*[lindex [subst \$label${suffix}_pos] 1]}]
+				set lx [lindex [subst \$label${suffix}_pos] 0]
+				set ly [lindex [subst \$label${suffix}_pos] 1]
+				if { $lx >= 0.0 && $lx <= 1.0} {
+					set "xlabel$suffix" [expr {$x+int($x1-$x)*$lx}]
+				} else {
+					set "xlabel$suffix" [expr {$x+$lx}]
+				}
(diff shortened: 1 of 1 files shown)


decentjohn:
Ok, can you make a decision whether to merge, by Sunday?  I'd like to give the beta a week, but no more if we don't get any feedback on bugs.

ebengoechea:
I thought new PRs and merges on the main branch were only for Nightly, isn't that the case? None of these changes are strictly needed in beta.

This is mergeable for me, thought you may want to test whether the -press* arguments work for you and that no side effects on Streamline. And I would publish it on Nightly.

decentjohn:
ah, if they're not needed for beta stability, then should we wait a week, until we can move the BETA to stable?

I'd rather hold off on any changes to nightly right now, only applying bug fixes we hear about it beta/nightly, testing them in nightly and then pushing them quickly to beta.

decentjohn:
maybe you're right, I'm not great on git.  I tagged it as beta, and was thinking "if we find bugs in the beta we should first fix them in the nightly, test, and the push to the beta"

ebengoechea:
Merged this for testing in Nightly now that Stable has been released.

PR #314, opened 2024-10-31
Update bluetooth.tcl

jianhaozh:
Add support for  LSJ-001 scale at Adonis's request.


jianhaozh:
this scale is  use the same protocol  as eureka  from Krell  but with different model name, so just add a new  if-else branch  to enter but use the same parser and other methods

decentjohn (approved):

PR #317, opened 2024-12-16
Add A_Flow plugin

Jan3kJ:
A_Flow is profile plugin based on D-Flow. It's a mix of D-Flow and Adaptive (for medium roasts) profile plus some adaptions.

I'm using it since beginning of the year and want to share it now.


decentjohn:
Super cool! I see a discussion of A_Flow here https://3.basecamp.com/3671212/buckets/7351439/messages/7332309713

I'm away from my DE1 this week, but will try it next week and if all looks good, approve it for the app.

PR #319, opened 2024-12-26
update a-flow to v1.1

Jan3kJ:
Update A-Flow to v1.1
changes: 
- updated UI and info text
- update readme
- remove not used buttons


decentjohn:
I can approve this, but for the next version can you make all the English text be translatable, by wrapping it like so:

[translate "always on"]		

Lots of people use the de1app in other languages.

Jan3kJ:
thanks. sure I will take care of this in the next version

Issue #316, opened 2024-12-13
streamline hot water temp and flow are wrong


diablodale:
Streamline skin shows wrong temp and wrong flow during hot water dispensing. Probably related to the tcl code below. It looks like the code is errantly displaying the same data for both dispensing hot water and flushing (aka rinse).

https://github.com/decentespresso/de1app/blob/0fc61ac4756cd3468d15ed055557531d47737a90/de1plus/skins/Streamline/skin.tcl#L733-L746

## Setup

* de1pro 1.45
* de1app 1.45.0.4

## Repo

1. App, calibrate, set hot water flow rate=10.0, set flush flow rate=6.0
2. Click OK and return to main streamline "page".
3. At left, tap hot water, tap temp, set to 75c
4. tap GHC to dispense hot water
5. Watch the flow and temp displayed on the hot water page

## Result

Hot water temp is incorrect 92c.
Hot water flow is incorrect 6ml/s.
They seem to be showing values (and the tcl code supports this) related to flush temp+flow.

![image](https://github.com/user-attachments/assets/dc227cfe-74d8-4d62-b1ba-55ddc263527b)
![image](https://github.com/user-attachments/assets/898df85d-5a1c-4cb6-a170-21976f15823d)
![image](https://github.com/user-attachments/assets/a495d273-5f81-46d1-9187-0355c544b42f)

## Expected

Hot water temp = 75c
Hot water flow = 10 ml/s

decentjohn:
Thanks, will check it out.

decentjohn:
Sorry for the delay in looking at this.

You pointed out a problem, but I think there's a different problem. 

Namely, it doesn't seem to me like setting "hot water temperature" setting actually works, even in Insight. Insight doesn't show the actual temperature of the hot water coming out, whereas Streamline does, so perhaps this was not noticed.    

Hot water appears to always use the temperature of the first step of Espresso.

Here's how I tested:

- set the hot water temp to 11C
- put my finger under the hot water output
- it feels very hot

But yes, you were also right that the Setting for hot water was being displayed, not the actual output. I've fixed that. https://github.com/decentespresso/de1app/commit/ccac4a7ba66b6999914b5d7165f95d942ddb8082

Could you do your own test on the hot water temperature, to confirm for me what I think I'm seeing?  I could very well be confused and fooling myself.

decentjohn:
[...]

PR #326, opened 2025-04-25
Separate hot water from flush buttons in Streamline skin

ufergus:
This commit splits the hot water buttons from the flush buttons in the Streamline skin.  The settings displayed with a common button list did not apply to hot water dispensing.  I think I found the appropriate hot water settings, but let me know if i need to tweak something to match Streamline goals.  Thanks.

Solves issue #316

### DIFF
--- a/de1plus/version.tcl
+++ b/de1plus/version.tcl
@@ -1,4 +1,4 @@
 package ifneeded de1app 99.99.99 {}
 # The package version is set by the build script. This is just a dummy value
 
-set ::changelog_link "https://decentespresso.com/de1app-v145.html"
+set ::changelog_link "https://decentespresso.com/blog/whats_new_in_de1app_v146"


ufergus:
Added another commit to add hot water dispense weight when bluetooth scale is connected.  Didn't mean for these to be merged into a single pull request.

decentjohn:
I'm sorry, but this button change is a design change, not necessarily something I'd accept as applying to all users, without the designers (Pulak) input.

Can you send me images of the changes you propose, and why?

In general, if you want to make a functional change to Streamline, please *first* bring up your idea for discussion.  Pulak has to be involved at all stages so that the UI remains clean and in his vision.

ufergus:
Thanks John.  I opened a thread in the Diaspora Programmers Forum to discuss further.

decentjohn:
Much appreciated, talk to you there.

PR #332, opened 2025-08-09
Fix the scale weight misreporting and remove unused code

### LINKED ISSUE #331: Weight flow rate is too high with DiFluid Microbalance
I'm investigating https://visualizer.coffee/shots/3dd1378d-6d80-45e1-b1a1-37f6b0381f4e which was recorded with a DiFluid Microbalance.

The puck seems to have reached full saturation between 17s and 23s, i.e. we achieve roughly stable weight flow rate for a stable input flow rate. The weight flow rate is reported at about 3g/s (+/- 10%), but the input flow rate is set to 1.7ml/s.

If we look at the measured weights, we have measured 22.1g at 17s and 34.6g at 23s. Calculating this we get: (34.6-22.10)/(23-17) ~ 2g/s - significantly closer to the input flow rate and 1g/s below the reported weight flow rate.

It seems to me that we do not utilize the weight flow rate reported by the scale, only it's weight, so it seems like the calculations are wrong somewhere.

Shot file:

[20250808T193727.json](https://github.com/user-attachments/files/21690039/20250808T193727.json)
[20250808T193727.shot.txt](https://github.com/user-attachments/files/21690170/20250808T193727.shot.txt)

As a rough estimation I reindexed the shot into 1s windows and calculated the differences between the 1s windows in this notebook:

https://gist.github.com/julian-klode/d92c8c066ced21ee9117acd108d95a73

Which confirms it's about 1.8-2g/s

julian-klode:
Use the actual time to calculate weight flow and remove the unused LSLR and period estimation code.

Fixes #331

Shot pulled: https://visualizer.coffee/shots/a00932fd-baf2-4a58-93c5-bad425ee108c

CC @jeffsf

### DIFF
--- a/de1plus/device_scale.tcl
+++ b/de1plus/device_scale.tcl
@@ -122,7 +122,6 @@ namespace eval ::device::scale {
 
 	proc init {} {
 
-		period::init
 		history::init
 	}
 
@@ -209,7 +208,6 @@ namespace eval ::device::scale {
 
 		# Update the internal registers
 
-		::device::scale::period::estimate_update $event_time
 		::device::scale::history::push_mass $reported_weight $event_time
 
 		set ::device::scale::_last_weight_update_time $event_time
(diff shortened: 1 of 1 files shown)


julian-klode:
We could limit this to the microbalance, and keep the period code for the others that may have stable periods. A problem with the period estimation is that it will only lower the estimate, and the Microbalance reporting intervals vary significantly, hence why the lowest value is always off (my reference shot was 0.166-1.766s with a mean of 0.249s and a standard deviation of 0.169s...).

So we could drop the Remove period commit, and pick the divisor passed on scale type (period * intervals for most scales, time difference for difluid). I don't know if it's necessary.

julian-klode:
Oh I think I accidentally ran this shit with high vibration filtering using the median estimator instead, this is a bit silly. Gonna edit the settings.tdb and rerun this in a couple of hours for an afternoon shot 😂

julian-klode:
Having simulated the original shot file from the bug after removing the calculated flow from it :) I still get differences vs smoothing over a window of 10 in the notebook, but certainly it's closer now :)

image

decentjohn:
The maths are beyond my ability to assess, but I do have one comment about estimating arrival time.  

Bluetooth guarantees package arrival order and packets to be delivered, so the weight measurements should be a 10hz basis, even if they don't arrive at that rate.  Could you comment on that?  Is the DiFluid NOT sending weight on a fixed-10hz basis (with the jitter being BLE comms?) ie is the not-10hz actually a firmware problem for the difluid.

Overall, I'm friendly to your change as the Half Decent Scale supports USB weighing, and thus faster-than-10hz is totally doable.   And Bengle will be USB based weighing as well, so faster-than-10hz is possible there too.

After you comment on how (if?) your changes would help/change scales that don't have any 10hz sending reliability issues?

julian-klode:
I'll abandon this PR for now after discussing this more with Jeff in #331 , because this introduced jitter due to Bluetooth jitter.

The nominal reporting interval for my scale seems to be 220 ms, but it's running slow and only reporting every 225 ms. I don't know if that's typical for the scale type.

For the Half Decent Scale and Bengle your life is a lot easier tbh: All you need to do there if you want it to report at say 100 Hz over USB, you just have the connect handler set the initial estimate to 0.01s instead of 0.1s.

You still want to keep the estimator likely to compensate for inconsistencies in the scales internal clock speed changing the sampling rate (as we see for my Microbalance).

Optimally HDS should probably do the estimator setup in the handshake: It could tell you its rate, then you don't need to hardcore it in de1app.

julian-klode:
Closing this will open a new one later if I get to a decent solution :)

PR #348, opened 2026-05-05
Protect file writes against power-loss corruption

voidptr:
# Protect file writes against power-loss corruption

## Problem

The app writes settings (and profiles, shots, etc.) by truncating the file and writing new content directly. If power is lost mid-write, the file is left empty or partially written. On next boot, `load_settings` reads the corrupt file and either crashes or silently loses all user settings.

This is a known issue on battery-free tablets that lose power unexpectedly.

## Solution

All file writes already funnel through a single `write_file` proc. This PR rewrites that proc to use an atomic write pattern:

1. Write data to `<filename>.tmp`
2. Copy existing file to `<filename>.bak` (original stays in place)
3. Rename `.tmp` over the target (atomic on POSIX/Android)

If power is lost at any step, the original file remains intact. One `.bak` generation is always retained.

Additionally, `load_settings` now detects corruption (empty file or malformed content) and automatically recovers from the `.bak` file. If both are bad, it falls through to the existing fresh-defaults path (same as a clean install). A toast notifies the user when recovery occurs.

A minor hardening: `translation.tcl` loading is wrapped in `catch` so a corrupt translation file degrades to English-only instead of crashing the app.

## Changes

| File | Change |
|------|--------|
| `safe_write.tcl` | NEW — `fast_write_open` + `write_file` with safe write-to-tmp → backup → rename pattern |
| `safe_load.tcl` | NEW — `load_settings_recover` proc: corruption detection + `.bak` recovery |
| `updater.tcl` | Sources `safe_write.tcl` (procs extracted for testability) |
| `utils.tcl` | Sources `safe_load.tcl`, calls `load_settings_recover` |
| `machine.tcl` | `translation.tcl` loading wrapped in `catch` |

## Scope

- All callers of `write_file` benefit automatically — no per-caller changes needed
- `write_file` signature is unchanged
- `append_file` is unchanged (appends don't truncate, so risk is only a partial trailing entry)
- Clean-install path (file doesn't exist) is unchanged
- `.tdb` format is unchanged

## Testing

The second commit adds a test suite (`tests/test_safe_write.tcl`) using Tcl's standard
`tcltest` package. 13 test cases covering normal writes, backup
creation, failure recovery, corruption detection, and the translation catch wrapper.
**This commit is optional** and can be dropped if you'd prefer not to introduce a test directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,8 @@ de1plus/plugins/log_upload/settings.tdb
 /de1plus/tmp/
 /de1plus/profiles_v2/*.json
 /de1plus/history_v2/*.json
+
+# Test artifacts
+/de1plus/tests/test_results.txt
+/de1plus/tests/test_error.txt
+/de1plus/tests/tmp_test_data/
--- a/de1plus/machine.tcl
+++ b/de1plus/machine.tcl
@@ -654,7 +654,10 @@ array set ::de1_substate_type_description {
 
 array set ::de1_substate_types_reversed [reverse_array ::de1_substate_types]
 
-array set translation [encoding convertfrom utf-8 [read_binary_file "[homedir]/translation.tcl"]]
+if {[catch {array set translation [encoding convertfrom utf-8 [read_binary_file "[homedir]/translation.tcl"]]}]} {
+    msg -ERROR "translation.tcl is corrupted or unreadable — using empty translations"
+    array set translation {}
+}
 
 proc de1_substate_text {} {
 	set num $::de1(substate)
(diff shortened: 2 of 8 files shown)


johnbuckman:
Could you test that the cloud update app still works even after your changes? 

The reason I ask is that it uses a minimum number of Tcl libraries, to try to ensure that syntax errors or other flaws in the program don't stop the updater from working. 

The absolute worst-case scenario is if someone commits a code change that breaks the updater, then people update and are unable to update thereafter because the updater itself is now broken. 

What this means is I'm not super happy changing things in the `updater.tcl` script because this has been battle tested to be known to work. I am really careful when/if I modify anything in it.

PR #352, opened 2026-07-27
Retire obsolete Decent Scale Off plugin

ODevStudio:
## Summary

- retire `Decent Scale Off` as an inert compatibility shim
- hide the obsolete plugin from the plugin list
- migrate persisted `decentscale_off` enablement out of settings
- add regression coverage for the shim, visibility, and migration

## Root cause

The plugin replaced the core `scale_disable_lcd` procedure and delayed the scale power-off write by three seconds. The normal sleep path disconnects the scale after one second, so the connection could close before the plugin sent the command. Core app code now owns this behavior and queues power-off before disconnecting.

## Impact

Existing installations receive the inert shim through the updater, while new users cannot enable the obsolete plugin. Persisted enablement is removed and saved once, preventing an older plugin copy from reactivating after a channel downgrade. The core `keep_scale_on` behavior remains unchanged.

## Validation

- `tests/decentscale_off.test.tcl`: 3 passed
- migration removes only known retired plugins and is idempotent
- `git diff --check`
- verified the core sleep path orders `scale_disable_lcd` before the delayed disconnect
- hardware validation not run

### DIFF
--- a/de1plus/plugins.tcl
+++ b/de1plus/plugins.tcl
@@ -197,6 +197,8 @@ namespace eval ::plugins {
             if {[string tolower $fbasename] == "dpx_steam_stop"} {
                 # incommpatible with firmware v1330 and newer because this functionality is already in the firmware, so plugin not needed
                 continue
+            } elseif {[string tolower $fbasename] == "decentscale_off"} {
+                continue
             } elseif {[string tolower $fbasename] == "skip_first_step_notice"} {
                 # per Damian's suggestion, should not be part of STABLE
                 continue
--- a/de1plus/plugins/decentscale_off/plugin.tcl
+++ b/de1plus/plugins/decentscale_off/plugin.tcl
@@ -1,43 +1,9 @@
-package require de1_bluetooth 1.1
-
 set plugin_name "decentscale_off"
-
 
 namespace eval ::plugins::${plugin_name} {
     variable author "Martin D"
     variable contact "Via Diaspora"
-    variable version 1.0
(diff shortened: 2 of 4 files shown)


johnbuckman:
Thank you for this — the diagnosis was right, and so was the point about persisted enablement, which turned out to be the part that actually matters.

We've retired the plugin in e7101e69, with a smaller change than this PR:

- `plugins.tcl`: `decentscale_off` joins the skip list, alongside the three other retired extensions.
- `utils.tcl`: one more `lsearch -inline -not -exact` line in the existing removal list in `load_settings`.

Both are needed, exactly as you reasoned: `plugins init` peeks `[plugins list]` but loads straight from `enabled_plugins`, so the skip list alone would leave the plugin loading for everyone who already had it switched on.

What we didn't take, and why:

- **`plugins/decentscale_off/plugin.tcl` is left untouched.** John's preference is not to gut it. Once it is neither listed nor in `enabled_plugins` it is never sourced, so there is nothing in it left to neutralise.
- **No `remove_retired_plugins_from_settings` proc and no `save_settings`.** The three existing retirements are plain filter lines that run on every `load_settings`, so a persisted entry is dropped on each launch; we kept that convention rather than adding a save.
- **The test then no longer applies**, since it covers the shim and the migration proc.

Also confirmed independently while reviewing: the core sleep path really does order `scale_disable_lcd` before the delayed `scale_disconnect_now`, so your root-cause description matches the code. Closing this in favour of the smaller change — much appreciated.

PR #357, opened 2026-08-25
Detect Bengle hardware from the model MMR and select the BLE protocol version

ChampionDesigns:
## Summary

Bengle support already exists here — `is_bengle_model`, the `BENGLE` scan-name match, and
the Bengle gate on `set_cupwarmer_temperature`, all added in `ee1a5da`. This PR replaces
the detection behind it and adds the protocol selector the rest of the Bengle series
depends on. Behaviour on a DE1 is unchanged apart from two scan-list bug fixes.

## Why the existing detection is not enough

`is_bengle_model` reads the BLE-advertised name, which is persisted to `settings.tdb`.
That breaks two ways:

- **Auto-reconnect never rescans.** `::de1_device_list` is empty, so `de1_ble` falls back
  to the literal `"DE1"` and overwrites the persisted value. A Bengle then reads as a DE1
  until someone runs a manual scan.
- **A stale name survives a firmware change.** `append_to_de1_list` returns early on a
  duplicate address, so a scan cannot correct a name already seeded from settings.

The model number has neither problem. It arrives from the `v13Model` MMR at `0x0080000C`
on every connect, from the machine.

## Changes

**`de1_de1.tcl`** — `is_bengle_model_value` (model **128 and above**, a range not a single
value, matching `isBengleModelValue` in `decentespresso/decaid`),
`detect_ble_protocol_version`, `set_ble_protocol_version`, `use_ble_v2`, and a write trace
on `::settings(machine_model)` so every connect path is covered without editing each one.

**`de1_comms.tcl`** — `is_bengle_model` returns `use_ble_v2`. `append_to_de1_list`
replaces rather than skips a duplicate address. Adds `long32_to_little_endian_hex`, a
32-bit companion to the existing 16-bit helper.

**`bluetooth.tcl`** — a connect with no scan entry keeps the persisted model instead of
overwriting it with `"DE1"`.

**`machine.tcl`** — `ble_protocol_version 1`.

**`vars.tcl`** — model 128 and above displays as `Bengle`.

**`skins/default/de1_skin_settings.tcl`** — the flow chart axes, their tick marks and two
`settings_2c2` editors follow `max_flowrate_v11` instead of a hardcoded 8 or 10.

## Flow rate

`::de1(max_flowrate_v11)` goes 8 → 20 mL/s on a Bengle and stays 8 on a DE1. That is the
live ceiling: `vars.tcl` range-checks three settings against it and the default skin reads
it for 16 slider and editor bounds, so the UI follows the machine.

The `_v11` name is now a misnomer. Renaming it would touch the default skin in 16 places
and is left for a separate change. `::de1(max_flowrate)` is untouched; nothing live reads
it.

## Why `use_ble_v2` has no caller here

It is the predicate the Bengle follow-ups gate on, and it is in this PR because the
detection that sets it is here. The same applies to `long32_to_little_endian_hex`: LED
colours are 24-bit, the machine clock runs to 604800, and the schedule word packs the day
into bits 31:22, so four later PRs need it. Defining it once here avoids four conflicting
copies.

## Test plan

- [x] `is_bengle_model_value` returns 0 for 1, 127, empty and non-numeric input, and 1 for
      128 and 129 — run in `tclsh`.
- [x] The write trace flips both the protocol and the flow ceiling in both directions.
- [ ] Connect a DE1. Confirm the info page names the correct model and the flow sliders
      still stop at 8 mL/s.
- [ ] TODO (Ben): connect a Bengle. Confirm the info page reads `Bengle` and the sliders
      reach 20 mL/s.
- [ ] TODO (Ben): restart the tablet and let it auto-reconnect with no scan. Confirm the
      machine is still identified correctly.

## Testing

Manual verification plus the two `tclsh` checks above. There is no automated test coverage
for these Tcl paths in the repo.

### DIFF
--- a/de1plus/bluetooth.tcl
+++ b/de1plus/bluetooth.tcl
@@ -2594,14 +2594,26 @@ proc de1_ble_handler { event data } {
 						::de1::event::apply::on_connect_callbacks \
 							[dict create event_time [expr {[clock milliseconds] / 1000.0}]]
 
+						# Look up the advertised name from the scan list. The list
+						# is runtime-only and empty on auto-reconnect (no fresh
+						# scan), so we only overwrite ::settings(model) when a
+						# real match is found. Otherwise whatever was persisted
+						# (e.g. "Bengle" from a prior scan) is preserved instead
+						# of being clobbered back to the default "DE1".
 						set espresso_machine_name "DE1"
+						set scan_name_found 0
 						foreach { entry } $::de1_device_list {
-							if { [dict get $entry address] eq $address} {								
+							if { [dict get $entry address] eq $address} {
 								set espresso_machine_name [dict get $entry name]
+								set scan_name_found 1
 								msg -INFO "Espresso machine model found: '$espresso_machine_name'"
 							}
 						}
-						set ::settings(model) $espresso_machine_name
+						if {$scan_name_found} {
+							set ::settings(model) $espresso_machine_name
+						} else {
+							msg -INFO "No scan entry for $address; preserving persisted model='[ifexists ::settings(model)]'"
+						}
 
 						de1_connect_handler $handle $address $espresso_machine_name
 
--- a/de1plus/de1_comms.tcl
+++ b/de1plus/de1_comms.tcl
@@ -2,6 +2,7 @@ package provide de1_comms 1.1
 
 package require de1_bluetooth
 package require de1_logging 1.2
+package require lambda
 
 ### Globals
 set ::failed_attempt_count_connecting_to_de1 0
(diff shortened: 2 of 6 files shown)


johnbuckman:
Only thing I disagreed with is:

> ::de1(max_flowrate_v11) goes 8 → 20 mL/s on a Bengle and stays 8 on a DE1. That is the
> live ceiling: vars.tcl range-checks three settings against it and the default skin reads
> it for 16 slider and editor bounds, so the UI follows the machine.

We can discuss, but my argument is that a DE1 owner shouldn't auto-modify a profile from 20mls to 8mls just because they can't do it.  If they did this, then a Bengle authored profile that then was shared to a DE1 and then was shared to a Bengle, would have been auto modified to 8mls. 

My feeling is that we never actually could guarantee 8mls, and now that we have a machine that can deliver it, we can let it be set.

However, I will mod the de1app to issue a warning that "Note: this model cannot achieve more than 8mls" if they set it higher.

johnbuckman:
Thanks Ben — this is landed in `main` as `0ec27a21` (verified on real DE1 and Bengle hardware: model detection, protocol v1/v2 flip, DE1↔Bengle cross-switch, cold auto-reconnect, and a normal DE1 shot with no regression).

Because the branch is on your fork and `machine.tcl` had drifted (`max_flowrate_v11` `8`→`20` in `fdd091f3`), it couldn't merge via the button, so it was carried into `main` directly with two adjustments — happy to discuss either:

1. **Uniform 20 mL/s ceiling for all machines** (per the discussion above): `max_flowrate_v11` is no longer switched by protocol — `set_ble_protocol_version` only selects the wire format, and the ceiling stays 20 for both DE1 and Bengle. This keeps a DE1 from auto-clamping a profile 20→8 (and a Bengle-authored profile survives a round-trip through a DE1). A DE1's v1 `U8P4` still physically caps at ~15.9; a toast warning for >8 mL/s on a DE1 will follow separately.
2. **`normalize_bengle_display_name`**: a Bengle whose firmware still advertises "DE1" over BLE was showing as "DE1" in the connect list — corrected from the authoritative model MMR.

Everything else is your PR as-is. Closing this in favour of the merged commit. **The six Bengle feature PRs can now re-file against `main`** — that's the next piece we're picking up. Thanks again!

